Issue #137, opened 2026-07-29
Proposal: new authkestra-devsig crate — device-bound request-signature verification (no session/user store)


stephane-segning:
## Summary

We'd like to propose a new crate, `authkestra-devsig`, implementing a per-request, device-bound signature authentication method — verification that needs **no session store, no token introspection, and no per-request network call**. A companion issue proposes the issuer-side half (attestation minting in `authkestra-op`); this one is the verifier.

We're ADORSYS-GIS / Webank, a mobile bank in Cameroon, replacing per-user OAuth tokens in our mobile client with this scheme in production. We cloned this repo and read the actual trait/module code (not just the README/book) before writing this, specifically to avoid proposing something that doesn't fit how the framework is actually wired today — some of what we found changes our own assumptions, and we've flagged that below rather than glossing over it.

We're not opening a PR yet — no merge rights, and we'd rather validate the approach with you first, especially the open design question in the "extension point" section below, since it affects the trait this crate implements.

## Background: the method, self-contained

Not an OAuth flow, not session-cookie auth, not bearer-token auth. Every request carries two JWS artifacts:

- **Request signature** — compact JWS, **detached payload** (RFC 7515 Appendix F: payload segment empty, signed content reconstructed by the verifier from the request itself). Protected header carries `alg: ES256` and the device's public JWK inline. Payload claims: `iat`, `exp` (60–120s), `jti`, `mth` (method), `pth` (path), `qsh` (query hash, if present), `bdh` (SHA-256 hex of raw body, if present), `aud`.
- **Attestation** — long-lived JWS from the Issuer (`authkestra-op`, see companion issue), binding `cnf.jkt` (RFC 7638 thumbprint of the device's public JWK) to `sub`, `did`, and application attributes.

Verification algorithm (normative order — cheap checks first, binding check is not skippable):

```
1. PRESENCE   both X-Signature and X-Attestation headers present, else REJECT
2. PARSE      both parse as compact JWS; alg ∈ allowed_algs (never "none", never symmetric)
3. ATTESTATION TRUST
              iss ∈ trusted_issuers; kid found in cached JWKS; signature verifies;
              not expired (± skew); att.status == "active"
4. THE BINDING   <- the security-critical step
              jwk ← sig.header.jwk; is a valid public key; contains NO private components (d, p, q, ...)
              thumbprint ← RFC7638_SHA256(jwk)
              constant_time_eq(thumbprint, attestation.cnf.jkt), else REJECT key_not_bound
[...]

Issue #136, opened 2026-07-29
Proposal: device attestation minting + enrolment PoP ceremony in authkestra-op


stephane-segning:
## Summary

We'd like to propose (and, if you're open to it, contribute) an extension to `authkestra-op` that lets it issue **device attestations** — a durable, signed binding of a client-held key to a user identity, used as one half of a request-signature authentication scheme (the other half, a verifier crate, is proposed separately in a companion issue).

This comes out of real production work: we're a mobile bank (ADORSYS-GIS / Webank, Cameroon) replacing per-user OAuth access/refresh tokens in our mobile client with device-bound signatures, and `authkestra-op` is structurally the right place to issue the credential because it already owns signing-key management, rotation, and JWKS publication.

We're opening this as an issue rather than a PR first because we don't have merge rights on this repo and want to validate the approach before writing code that might diverge from what you'd actually want merged.

## Background: what an "attestation" is and why it needs an issuer

(Self-contained summary — you don't need any other context to follow this.)

Every request from our client carries two JWS artifacts that do different jobs:

- a short-lived, single-use **request signature**, signed by a key that lives only on the device (never transmitted), whose protected header embeds the device's public JWK; and
- a long-lived, reusable **attestation**, signed by an Issuer, binding that key's RFC 7638 thumbprint to a user identity and attributes.

Neither is sufficient alone: a self-signed key proves possession of *a* key, not entitlement to *an account*. The `key → identity` binding has to live somewhere, and the only arrangement that's both identified and requires no per-request database lookup is one where a trusted Issuer signs that binding and it travels with the request. That's the attestation, and `authkestra-op` — already an OpenID Provider — is the natural Issuer, because it already has everything structurally required except the issuance call itself.

Attestation claim shape (`RS256`, verifiable via the OP's existing `/jwks.json`):

```json
{
  "iss": "https://id.example.com",
  "sub": "user-7f3a9c12",
  "cnf": { "jkt": "0ZcOCORZNYy-DWpqq30jZyJGHTN0d2HglBV3uiguA4I" },
  "did": "dev-8821bc04",
  "iat": 1785225600,
  "exp": 1787817600,
  "att": { "kyc_level": 2, "roles": ["customer"], "status": "active" }
}
```

[...]

PR #138, opened 2026-07-30
feat(devsig): device-bound signature authentication (authkestra-devsig)

stephane-segning:
## Summary

Adds `authkestra-devsig`, a new crate implementing **device-bound signature authentication**:
per-request proof of possession of a device-bound private key (`X-Signature`), plus a long-lived
Issuer-attested key binding (`X-Attestation`). Verification needs no session store, no token
introspection, and no per-request network call — any service holding one cached Issuer JWKS can
verify a request independently.

Closes / addresses #137.

## Why

A mobile client whose private key never leaves hardware-backed storage (Secure Enclave /
StrongBox / Keystore) doesn't need OAuth-style refresh cycles — every request is its own proof.
This is the same idea as WebAuthn/DPoP/mTLS, applied as a first-class request-authentication
method: the signature proves *possession of a key*, and the attestation (signed by an Issuer,
e.g. `authkestra-op`) proves *whose key it is*. Neither is sufficient alone — see "The
security-critical check" below.

## What's in this PR

- `verify()` — a plain, framework-agnostic async function implementing the full algorithm in
  order: presence → parse + `alg` check (both credentials) → attestation trust (issuer, `kid`,
  signature, expiry, status) → **the key-thumbprint binding check** → request-signature
  verification → freshness (skew + max lifetime) → request binding (method/path/audience/query
  hash/body hash) → replay (fails closed on any store error).
- `VerifierConfig`, `IssuerJwks` (a minimal in-memory Issuer-key cache — see `src/jwks.rs` for
  why `authkestra_resource::jwt::JwksCache` isn't a drop-in fit here), `ReplayStore` trait +
  in-memory impl, `DeviceIdentity` (the output), `VerifyError` (one variant per named rejection).
- An optional `axum` feature: a `tower::Layer` that buffers the body (size-capped), calls
  `verify()`, and injects a verified `DeviceIdentity` into request extensions, plus a thin
  `FromRequestParts` extractor that reads it back out. See "Integration surface" below — **this
  part is a proposal, not a settled decision.**
- 24 conformance tests (`tests/conformance.rs`) covering presence, `alg` rejection (including
  `none` and HS256 specifically, not just "any error"), the binding check, body/query/method/path
  tampering, replay, fail-closed replay-store-unavailable, clock skew, and a private-key-in-`jwk`
  smuggling attempt.

## The security-critical check

An attacker who has captured a victim's attestation (it's public — it travels in every request)
can pair it with a request signed by their *own*, genuinely-held key. The attestation's signature
verifies fine (it's a real Issuer signature). The request signature verifies fine (against a key
the attacker really holds). **Only comparing the embedded `jwk`'s RFC 7638 thumbprint against the
attestation's `cnf.jkt` catches this** — `case_04_attestation_for_a_plus_attackers_key_is_rejected_key_not_bound`
in `tests/conformance.rs`.

I verified this test is load-bearing, not just present: I temporarily replaced the check
(`if !constant_time_eq(...)` → `if false && !constant_time_eq(...)`) and reran just that test.

**Before (check disabled):**
```
test case_04_attestation_for_a_plus_attackers_key_is_rejected_key_not_bound ... FAILED
thread '...' panicked: called `Result::unwrap_err()` on an `Ok` value: DeviceIdentity { subject:
"user-7f3a9c12", device: "dev-8821bc04", ... }
```
i.e. the attacker's key was silently accepted as the victim's identity — a total authentication
bypass, exactly as expected once the check is removed.

**After (fix restored):** all 25 tests pass (see `cargo test` output below).

## A real bug found while implementing this (not part of the original ask)

`jsonwebtoken` 10.4.0 — the version already pinned in this workspace — has
`Jwk::thumbprint()` **panic** (not return a `Result`) on a structurally-inconsistent `kty`/`crv`
pairing, e.g. `{"kty":"EC","crv":"Ed25519"}`. `EllipticCurveKeyParameters` and
`OctetKeyPairParameters` share the same `EllipticCurve` enum for their `crv` field, so serde
happily deserializes that combination into a well-typed `Jwk` — the panic only fires inside
`thumbprint()`'s match arm. Since the `jwk` here comes straight from the **attacker-controlled**
`X-Signature` header (verified before it can be checked against `cnf.jkt`, by construction), a
single crafted request could otherwise panic the verifying task.

Fixed in `src/signature.rs` with two layers: an explicit `kty`/`crv` consistency check before
ever calling `thumbprint()` (rejects as `bad_jwk`), plus a `catch_unwind` around the call itself
as defense-in-depth against any other input this crate hasn't enumerated. Covered by
`inconsistent_curve_jwk_is_rejected_as_bad_jwk_without_panicking` in `tests/conformance.rs`,
which fails loudly (test process aborts) without the fix and passes cleanly with it. **Not
fixed by bumping `jsonwebtoken`** — that's a workspace-wide dependency decision out of scope for
a new-crate PR; this guards the call site instead.

## Integration surface — genuinely open, your call

Two extension points exist in `authkestra-engine` today, and I confirmed against the current
code (not assumed) that neither can express what this verifier needs:

- `AuthMethod` (`auth/mod.rs`) is roadmapped but used only in `tests.rs`; its `AuthInput` is
  `Password | OAuthCode | Token | Custom(json)` — no HTTP request context at all.
- `AuthenticationStrategy<I>` (`auth/strategy.rs`) is what `Guard<I>`/`JwtStrategy` actually
  chain (confirmed live via `authkestra-resource/src/lib.rs` and `jwt.rs`), but
  `authenticate(&self, parts: &Parts)` only ever sees `Parts` — no body. Every existing axum
  extractor in `authkestra-axum` implements `FromRequestParts`, not `FromRequest`, for the same
  reason.

This verifier needs the raw body for the `bdh` (body-hash) check, so it would be the framework's
first body-aware auth extractor regardless of which trait it targets — which is exactly the
question this issue raised. I did not resolve it: **the verification core (`verify()`) is
complete and tested entirely independently of this question.** For today's integration surface I
implemented the `tower::Layer` approach floated in this issue's comments, specifically *because*
it needs neither trait to exist yet. I deliberately did **not** implement
`AuthenticationStrategy<I>` for this verifier — a real impl of that trait would have exactly one
option (silently skip the `bdh` check whenever a body is present, because the trait can't see it),
and shipping that silently felt like exactly the kind of unilateral call this issue asked not to
make. Migrating to whichever trait you land on is then a matter of writing a thin adapter that
builds a `SignedRequest` from whatever the trait provides and calls the same `verify()` — nothing
in `verify.rs`, `signature.rs`, or `attestation.rs` needs to change for that migration to happen.

I'd genuinely like your input on #137's open question before the integration half of this is
considered final. Happy to rework `axum_integration.rs` however you'd prefer once you've had a
chance to think about it.

## What this PR deliberately does not do

- Does not touch `authkestra-op` (issue #136, the Issuer / attestation-minting side) — that's a
  separate PR. This crate only *verifies* attestations; it never mints them. Test attestations
  are self-signed with a test RSA key in `tests/support/mod.rs`.
- Does not upgrade `jsonwebtoken` or any other pinned dependency version.
- Does not implement Actix integration — only `axum`, gated behind a feature. Happy to add Actix
  if useful, but wanted to keep this PR's surface reviewable and not guess at demand.
- Does not add this crate to `.github/workflows/publish.yml`'s publish matrix — can follow once
  the design settles.

## Verification

```
$ cargo build -p authkestra-devsig --all-features
    Finished `dev` profile [unoptimized + debuginfo] target(s)

$ cargo test -p authkestra-devsig --all-features
running 25 tests
...
test result: ok. 25 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3.41s

$ cargo clippy -p authkestra-devsig --all-features --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s)   # zero warnings

$ cargo fmt -p authkestra-devsig -- --check   # clean

$ cargo build --workspace --all-features && cargo test --workspace --all-features \
  && cargo clippy --workspace --all-features -- -D warnings
# all green, nothing else in the workspace affected
```

I also independently confirmed this crate's public API is consumable end-to-end against a real
attestation minted by an in-progress `authkestra-op` enrolment implementation (issue #136, being
worked on in parallel) — an attestation minted there, fed into `authkestra_devsig::verify` here,
round-trips correctly. That's not part of this PR (it lives in the other branch), but it's real
evidence the two independently-implemented sides of the spec's wire format actually agree.

## AI usage disclosure

This implementation was produced with AI assistance (Claude), starting from a tested prototype
(`ADORSYS-GIS/webank-services`, `crates/devsig`) built against the same `jsonwebtoken` 10.4.0
pinned in this workspace, and adapted to this repo's crate layout and conventions. I read every
file in `authkestra-engine`/`authkestra-resource`/`authkestra-axum` referenced above directly
before writing this, rather than trusting secondhand descriptions — several of the claims above
(which trait is actually live, whether `JwksCache` is seedable, the `jsonwebtoken` panic) were
verified against the real source in this repo and the real vendored dependency source, not
assumed. All tests in this PR were run locally and are pasted above, not fabricated.


stephane-segning:
Suite au retour de @marcjazz (relayé par @stephane-segning) : ajout de l'intégration Actix, des exemples d'intégration, et mise à jour de la documentation. Résumé des changements poussés sur cette branche :

- **Rebase sur `develop`** : la branche a été mergée avec `upstream/develop` (seul conflit réel : la liste `members` du `Cargo.toml` racine, entre l'entrée `authkestra-devsig` de cette branche et la nouvelle entrée `examples/axum-mfa-server` de `develop` — résolu en gardant les deux). Le PR est de nouveau `mergeable: MERGEABLE` contre `develop`.
- **Intégration actix ajoutée** : `authkestra-actix/src/devsig.rs` fournit `DeviceSignatureAuth` (middleware `actix_web::dev::Transform`, équivalent du `tower::Layer` axum) et l'extracteur `AuthDeviceSignature`, activables via la feature `devsig`.
- **Repositionnement architectural de l'intégration axum** : conformément à la règle "Framework Agnostic" d'`AGENTS.md`, l'intégration axum (qui vivait dans `authkestra-devsig/src/axum_integration.rs`, à l'intérieur du crate de vérification) a été déplacée vers `authkestra-axum/src/devsig.rs`. `authkestra-devsig` reste désormais 100% agnostique du framework — même structure que `authkestra-op` → `authkestra-axum/src/op.rs` / `authkestra-actix/src/op.rs`.
- **Exemples d'intégration ajoutés** : `crates/authkestra-axum/examples/devsig/` et `crates/authkestra-actix/examples/devsig/` — chacun génère une paire de clés Issuer/device de démo au démarrage, protège une route `POST /v1/transfer`, et affiche une commande `curl` prête à l'emploi avec une requête signée valide. Testés manuellement : les deux répondent bien `200 OK` avec une requête signée valide et `401 Unauthorized` sans en-têtes.
- **Documentation mise à jour** : `docs/book/ch06-adapters-and-integrations.md` (nouvelle section devsig), ainsi que les README de `authkestra-devsig`, `authkestra-axum` et `authkestra-actix`.
- Chaque nouveau handler/middleware est instrumenté avec `tracing` (règle DoD d'`AGENTS.md`).

Vérifications locales : `cargo test --workspace --all-features` (toutes vertes, y compris les 25 tests de conformité devsig), `cargo clippy --workspace --all-features -- -D warnings` (clean), `cargo fmt --all -- --check` (clean), `cargo llvm-cov --workspace --all-features --fail-under-lines 50` (60,47% de couverture lignes, au-dessus du seuil).

marcjazz (approved):

PR #139, opened 2026-07-30
authkestra-op: device/service attestation issuance (enrolment + re-issuance) — closes #136

### LINKED ISSUE #136: Proposal: device attestation minting + enrolment PoP ceremony in authkestra-op

stephane-segning:
## Summary

Implements the Issuer side of the device-bound-signature authentication method in `authkestra-op`: the enrolment/re-issuance ceremony and `cnf.jkt`-bound attestation minting described in the design spec linked below. This is a proposal for discussion, not a finished, non-negotiable design — see "Open design choices" below for the two points I'd especially like feedback on.

Closes #136.

Source of truth: the full spec this implements is `authkestra-device-signature-spec.md` §5.6/§5.6.1 (the enrolment ceremony and the service-principal addition), written against this repo. Happy to paste the relevant sections inline if useful for review — let me know.

## What this adds

All new code lives in `authkestra-op` (framework-agnostic, like every other handler in this crate) plus thin route wiring in `authkestra-axum`/`authkestra-actix` per this repo's own `AGENTS.md` DoD ("every new handler added to `authkestra-op` MUST have a corresponding route wired in both adapters in the same PR").

- **`src/attestation.rs`** — the core types and crypto:
  - `PrincipalType` (`Device` | `Service`, spec §5.6.1)
  - `parse_public_jwk` — validates a caller-submitted JWK: rejects anything that isn't a JSON object, rejects any private/symmetric component (`d`/`p`/`q`/`dp`/`dq`/`qi`/`k`) checked against the **raw JSON before typed parsing** (`jsonwebtoken::jwk::Jwk` silently drops unknown members, so a post-parse check would miss a smuggled `d`), and rejects symmetric (`oct`) keys outright
  - `compute_cnf_jkt` — RFC 7638 thumbprint, always computed by the OP from a JWK it just verified possession of, **never accepted as caller input**
  - `EnrolmentChallenge` / `EnrolmentChallengeStore` — a blanket impl over any `KvStore<T> + AtomicConsume<T>` backend, copied directly from `AuthorizationCodeStore`'s existing pattern in `code.rs` (same single-use/short-TTL shape)
  - `SecondFactorVerifier` and `AttestationStatusProvider` — the two pluggable hooks (see "Open design choices")
- **`src/handlers/enrolment.rs`** — the request/response DTOs and three handlers:
  - `handle_enrol_start` — validates the JWK, calls the configured `SecondFactorVerifier`, issues a server-generated single-use challenge
  - `handle_reissue_start` — re-issuance path (see below)
  - `handle_complete_challenge` — shared by both flows: consumes the challenge, verifies the signature against the JWK submitted when the challenge was issued, computes `cnf.jkt` itself, mints the attestation
- **`authkestra-engine`**: one additive method, `TokenManager::issue_custom_token`, so the attestation can carry the wire-format's `typ: "webank-attest+jws"` header (the existing `issue_*_token*` family hardcodes `typ: "JWT"`). Existing methods are unchanged.
- **Routes** (both adapters): `POST /enrol`, `POST /enrol/complete`, `POST /reissue`.

Attestations are minted with the OP's existing `TokenManager` — the same signing key, rotation and JWKS publication `authkestra-op` already has for OIDC. No new key material, no new infrastructure.

## The security-critical property, tested explicitly

`cnf.jkt` is computed by the OP from the JWK it verified proof-of-possession over, and is never read from anything the caller submits. `caller_supplied_cnf_jkt_is_ignored_not_honoured` in `handlers/enrolment.rs` mints an attestation where the request supplied a `requested_cnf_jkt` different from the real key's thumbprint, and asserts the issued attestation's `cnf.jkt` equals the real key's RFC 7638 thumbprint — not the supplied value.

Other tests cover: a signature over the right challenge but from a *different* key (rejected), a challenge that was never actually issued (rejected — a client-chosen nonce doesn't satisfy PoP), challenge replay (rejected, single-use), `alg: none` on the challenge signature (rejected), a JWK carrying a smuggled private component (rejected), and the re-issuance binding check (below).

### 1. Second factor for service principals (spec §5.6.1, explicitly left open)

A device can receive an SMS OTP; a service can't. Rather than hardcode SMS OTP anywhere, or special-case services, I made this a pluggable trait:

```rust
#[async_trait]
pub trait SecondFactorVerifier: Send + Sync {
    async fn verify(&self, subject: &str, principal_type: PrincipalType, proof: &SecondFactorProof) -> Result<(), OpError>;
}
```

`SecondFactorProof { kind: String, value: String }` is opaque to `authkestra-op` — the host application supplies whatever verification makes sense per `PrincipalType` (SMS/TOTP for devices; an out-of-band admin approval or a one-time bootstrap secret consumed exactly once for services). This felt like the most idiomatic shape for a generic auth framework that shouldn't hardcode a telecom integration, but it's a real design call, not an obviously-correct one — alternatives I considered and didn't pick:
- Baking in an OTP-only implementation and telling service callers to fake one (rejected: leaks a device-shaped assumption into a service-shaped problem)
- A required bootstrap-secret mechanism built into this crate (rejected: forces one specific mechanism on every deployment; some may prefer an admin-approval queue instead)

Happy to go a different direction if you have one in mind.

### 2. Re-issuance mechanism (ADR 0014 decision point 6: 24h lifetime, silent re-issue at 12h)

The spec poses this as open: does re-issuance need the full ceremony every 24h, or can an about-to-expire attestation itself serve as proof for a lighter-weight re-attest? I implemented the latter:

`handle_reissue_start` takes the current (still valid) attestation plus a freshly-submitted public JWK. It validates the attestation, computes the submitted JWK's thumbprint, and compares it against the attestation's `cnf.jkt` — **the exact same binding check the request-signature verifier (`authkestra-devsig`, #137) performs on every ordinary request.** If it matches, a new challenge is issued (same PoP requirement as first enrolment — still no bare client-chosen nonce), and completing it mints a fresh attestation. No second factor is re-verified.

Reasoning: the second factor was already verified once, at initial enrolment; re-issuance's job is to prove *continuity* (still the same key), not re-establish identity from scratch. This also means a lost/rotated key **cannot** silently re-issue — it fails the binding check exactly like a forged pairing would.

For revocation and attribute freshness, I added an optional hook:

```rust
#[async_trait]
pub trait AttestationStatusProvider: Send + Sync {
    async fn current_attributes(&self, subject: &str, principal_id: &str, principal_type: PrincipalType) -> Result<Option<Value>, OpError>;
}
```

If configured, it's called on every re-issuance; `Ok(None)` refuses re-issuance (`PrincipalRevoked`) and `Ok(Some(attrs))` refreshes `att` with current data — so a device revoked by the host application stops being silently re-attested at the next 12h checkpoint rather than riding out the full 24h window on stale claims. If not configured, re-issuance falls back to copying the previous `att` claim forward unchanged (still bound by proof-of-possession, just blind to revocation between full ceremonies).

Alternative I didn't pick: requiring the full second-factor ceremony on every re-issuance. Rejected because it defeats "silent" — an SMS OTP every 12 hours is not silent, and ADR 0014 is explicit that the client-side retry path needs to work unattended.

Both of these are genuinely adjustable — if either doesn't fit how you want the framework to look, I'm glad to rework it.

## Verification

```
cargo build --workspace
cargo test --workspace
cargo clippy --workspace --all-targets -- -D warnings
cargo fmt --all -- --check
```

All green. `authkestra-op`'s own suite: 68 passed, 0 failed (`cargo test --workspace`) — 67 from the initial implementation plus 1 new regression test added during the review pass below (`--all-features` reports 72; the extra 4 are pre-existing feature-gated tests unrelated to this change, not new ones — checked precisely rather than assumed, since this PR is asking for extra scrutiny on exactly this kind of claim). Also spot-checked `authkestra-axum`/`authkestra-actix` compile and lint clean with `--features op,flow` (the minimal combination that exercises the new routes — see note below).

I additionally wrote (and ran locally, then removed before this push) an integration test that minted an attestation with this code and fed it, plus a hand-built request signature using the same device key, through `authkestra_devsig::verify` from #137's branch — confirming the two independently-implemented crates actually agree on the wire shape end-to-end, not just that each looks right in isolation. It passed. I didn't include it in this PR because `authkestra-devsig` isn't part of `main` yet (that's #137, still open) and a path dependency on a sibling in-flight PR would break this one's build for anyone reviewing it in isolation. Happy to add it as a workspace-level test once both land, if that's useful.

## A real vulnerability found and fixed in review, before this was pushed

`jsonwebtoken` 10.4.0 (this workspace's pin) makes `Jwk::thumbprint()` **panic**, not return `Err`, on a structurally-inconsistent key — e.g. `kty: "EC"` paired with `crv: "Ed25519"` (`panic!("EllipticCurve can't contain this curve type")`). `compute_cnf_jkt` calls this on a JWK that is attacker-controlled by construction (that's the entire premise of this ceremony), so an unguarded call is a crash-on-attacker-input DoS — and `handle_reissue_start` called it **before the presented attestation is even validated**, making `POST /reissue` reachable pre-auth.

Fixed once, in `compute_cnf_jkt` itself, rather than trusting every call site to remember a guard: it now wraps the call in `std::panic::catch_unwind` and returns `Result<String, OpError>`, rejecting as `bad_jwk` instead of unwinding. Added `compute_cnf_jkt_rejects_inconsistent_curve_without_panicking` as a regression test — this is exactly the shape of finding I'd want a reviewer to double-check independently rather than take on faith, so please do.

## Things I noticed along the way (not blocking, just flagging)

- `authkestra-op/src/lib.rs`'s doc comment still says "this crate is a skeleton... no handler logic has landed yet" — stale, given `handlers/` is ~3,400 lines including Device Authorization Grant and Token Exchange. Didn't touch it since it's unrelated to this change, but it might be worth a follow-up.
- Building `authkestra-axum` or `authkestra-actix` with `--features op` alone currently fails to compile (pre-existing, unrelated to this PR): `op`'s Cargo.toml feature only pulls in `session` + `token`, but the `AuthSession` extractor both crates already have needs `SessionConfig`/`FromRequest`, which is only available under `flow`. I verified my new code with `--features op,flow` instead. Worth its own fix if useful — let me know and I can split it out.
- The spec originally described `X-Signature` as a detached-payload JWS; that was corrected before I started this work (§3.2 of the spec has the note) after an earlier prototype caught that `iat`/`exp`/`jti` have no channel for the verifier to "reconstruct" them from. Not this PR's concern, but flagging since it's the same wire format this attestation needs to agree with downstream.

## AI usage disclosure

I used an AI coding assistant to explore this codebase, design against the spec, implement the ceremony, write the tests (including the security-critical negative cases), and verify the build/test/clippy/fmt output shown above. I reviewed the diff and reasoning myself before opening this PR and take responsibility for it; treat it as you would any other external contribution — happy to walk through any part of it in review.


stephane-segning:
## Suivi du retour du mainteneur (intégration actix, exemples, documentation)

J'ai vérifié ce que ce PR couvrait déjà avant de commencer, plutôt que de supposer :

**Intégration actix — déjà présente, vérifiée par lecture directe.** `crates/authkestra-axum/src/op.rs` et `crates/authkestra-actix/src/op.rs` câblent déjà les trois routes (`/enrol`, `/enrol/complete`, `/reissue`) avec instrumentation `tracing` dans les deux adaptateurs, conformément à la DoD du dépôt. Rien à ajouter là-dessus.

**Ce qui manquait réellement : exemples et documentation.** Ajoutés dans `af384d7` :

- `crates/authkestra/examples/axum/op_server_attestation.rs` et `crates/authkestra/examples/actix/op_server_attestation.rs` : exemples autonomes et exécutables (aucun service externe requis — store de challenge en mémoire) qui font tourner tout le cycle enrôlement + ré-émission contre une vraie instance `authkestra-op`, avec un vrai client générant une clé P-256. Les deux ont été exécutés manuellement de bout en bout ; le `cnf.jkt` de l'attestation initiale et celui de la ré-émission correspondent bien à la même clé.
- `docs/book/ch06-adapters-and-integrations.md` : nouvelle section « Device/Service Attestation Issuance (OP) » qui documente les trois routes, les deux hooks branchables (`SecondFactorVerifier`, `AttestationStatusProvider`) et la différence de câblage axum/actix. (`crates/authkestra-op/README.md` n'existe pas dans ce dépôt — rien à mettre à jour côté README de crate.)

**Un vrai bug trouvé en écrivant l'exemple Axum.** Les bounds de `OpExt::op_axum_router()` avaient grossi pour exiger inconditionnellement `EnrolmentChallengeStore`/`SecondFactorVerifier`/`AttestationStatusProvider`/`AttestationConfig` — ce qui casse la compilation de tout `AppState` existant qui n'a pas ces champs. Ce n'est pas hypothétique : ça cassait déjà la compilation des trois exemples OP préexistants du même crate (`op_server.rs`, `op_server_sqlx.rs`, `op_server_custom_grant.rs`). `cargo test --workspace` ne l'attrape pas car les exemples ne font pas partie des cibles de test par défaut. Corrigé en séparant les trois routes d'attestation dans une méthode dédiée `op_axum_attestation_router()`, composable via `.merge()` — cohérent avec la justification déjà donnée dans ce PR pour garder `AttestationConfig` séparé d'`OpConfig`. Côté actix, `op_actix_scope()` n'a pas ce problème : chaque dépendance est résolue individuellement via `app_data`, pas via un seul bound générique tout-ou-rien.

**Vérification** (dépôt cloné à partir de la branche, `develop` comme base) :
- `cargo test --workspace --all-features` : 125 tests passés, 0 échec
- `cargo clippy --workspace --all-features -- -D warnings` : propre
- `cargo fmt --all -- --check` : propre
- Les deux nouveaux exemples exécutés manuellement de bout en bout (pas seulement compilés)

Poussé sur `feat/device-attestation-enrolment-136` (pas de nouveau PR).

stephane-segning:
Bonjour @marcjazz,

Suite à ton retour ("missing website docs"), j'ai ajouté la documentation manquante sur le site public Astro/Starlight (`website/`), distincte du chapitre mdBook déjà présent dans cette PR :

**1. `website/src/content/docs/advanced/op-server.md`** — nouvelle section H2 « Device/Service Attestation Issuance » (même profondeur/style que les sections existantes du fichier : prose + extrait Rust montrant `AttestationConfig`, `SecondFactorVerifier`, et le wiring des routers). La liste des endpoints exposés dans « Wiring the OP Endpoints » a aussi été mise à jour pour inclure les trois nouvelles routes (`POST /enrol`, `POST /enrol/complete`, `POST /reissue`) à côté des quatre existantes.

**2. `website/src/content/docs/guides/device-attestation.md`** — nouveau guide pas-à-pas (ton tutoriel, comme `guides/wired-endpoints.md`, pas le ton référence de `op-server.md`) qui reprend exactement le déroulé des deux exemples exécutables déjà présents dans cette PR (`crates/authkestra/examples/{axum,actix}/op_server_attestation.rs`) : génération de la clé, `/enrol`, signature + `/enrol/complete`, puis `/reissue` avant expiration. Le guide référence explicitement ces deux fichiers d'exemple.

**3. Sidebar** — la nouvelle page est enregistrée dans `website/astro.config.mjs` sous le groupe **« Building Servers »**, à côté de la référence OP-server, plutôt que sous « Getting Started ». Raison : ce guide suppose un serveur `authkestra-op` déjà en place avec le router d'attestation câblé — c'est un approfondissement avancé et spécifique à l'OP server, pas une prise en main générale comme Quickstart/Framework Integration/Wired Endpoints.

**4. Conflit avec `develop`** — la branche avait un vrai conflit de fusion (un seul, réel) : la PR #138 (device-bound signature auth, `authkestra-devsig`) et cette PR avaient toutes les deux ajouté une nouvelle section H2 au même endroit dans `docs/book/ch06-adapters-and-integrations.md`. Résolu par un **merge** (pas un rebase, pour préserver l'historique de la branche comme pour #138) de `upstream/develop` : les deux sections H2 sont conservées (celle sur `authkestra-devsig` d'abord, puis celle sur l'attestation issuance), rien n'a été supprimé côté #138. Tout le reste du merge s'est fait automatiquement sans conflit.

**Vérifications effectuées** (sur mon clone, avant push) :
- `pnpm install --frozen-lockfile && pnpm run build` (site) → build propre, 25 pages générées dont `/guides/device-attestation/`, aucun warning.
- Confirmé que l'entrée sidebar « Device Attestation » apparaît bien sous « Building Servers » dans le HTML généré, et que l'ancre `#deviceservice-attestation-issuance` référencée depuis le guide correspond exactement au slug généré pour le titre H2.
- `cargo check --workspace --all-features --all-targets` → OK.
- `cargo test --workspace --all-features` → 152 tests passés, 0 échec.
- `cargo clippy --workspace --all-features -- -D warnings` → aucun warning.
- `cargo fmt --all -- --check` → OK.

Poussé sur `feat/device-attestation-enrolment-136` (pas de nouvelle PR). La PR est maintenant `mergeable: MERGEABLE` côté GitHub (elle était `CONFLICTING` avant le merge de `develop`).

marcjazz (changes requested):
I've reviewed the PR. Overall, the implementation correctly applies the spec and the `cnf.jkt` binding check is robust. However, there are a few issues that need to be addressed before this can be merged:

1. **Pre-auth DoS Vector in `handle_reissue_start`:** Even though you fixed the panic in `compute_cnf_jkt` using `catch_unwind`, you are still calling `parse_public_jwk` and `compute_cnf_jkt` **before** validating the presented attestation via `tokens.validate_token`. This allows any unauthenticated attacker to force the server to parse JSON and compute SHA-256 hashes by hammering the `/reissue` endpoint. You must flip the order: validate the attestation *first*, and only process the JWK if the attestation is valid.

2. **Feature Gate Compilation Issue:** You noted in the PR description that building `authkestra-axum` or `authkestra-actix` with `--features op` fails to compile because it lacks `flow` dependencies. This needs to be fixed in this PR. Either the `op` feature must enable `flow`, or the relevant code in the adapters needs to be feature-gated so that `--features op` compiles cleanly on its own. Breaking the build on valid feature combinations is not acceptable.

3. **Stale Documentation:** As you pointed out, `authkestra-op/src/lib.rs` still claims the crate is a skeleton with no handler logic. Since you are well aware of this, please include the fix in this PR rather than leaving it for a follow-up.

Regarding your open design choices:
- The `SecondFactorVerifier` trait design is excellent. It correctly avoids hardcoding SMS/OTP and perfectly aligns with the framework-agnostic typestate rules (by using `&dyn SecondFactorVerifier` avoiding generics).
- The silent re-issuance mechanism (ADR 0014 decision point 6) is solid. Relying on key continuity for the 12h re-issuance without prompting for a second factor is the right UX call, and the `AttestationStatusProvider` revocation hook is well-placed.

Please address the pre-auth parsing and feature compilation issues, then I'll be happy to approve.

marcjazz (approved):

PR #141, opened 2026-07-31
fix(op): pre-auth cost amplification in reissue, op-standalone build, stale doc

stephane-segning:
## Summary

Addresses all three items from your `CHANGES_REQUESTED` review on #139 (https://github.com/marcjazz/authkestra/pull/139#pullrequestreview-4820581745). Verified each was still present on `develop` (`224a7ce`) before fixing anything — the review and the merge were 30 minutes apart with no commit in between, so I checked rather than assumed.

1. **Pre-auth cost amplification in `handle_reissue_start`, still live.** `parse_public_jwk`/`compute_cnf_jkt` ran before `tokens.validate_token`. Reordered — attestation validated first, JWK only touched once it's known genuine.
2. **`--features op` still failed standalone on both adapters.** Reproduced both failures directly:
   - `authkestra-axum`: `AuthSession`'s `FromRequestParts` impl needs `SessionConfig` (only exported under `flow`) but was gated on `session` alone; `op` enables `session`+`token`, not `flow`. Fixed by scoping `AuthSession` itself to `all(session, flow)` — confirmed `op.rs` in this crate never references `AuthSession`, so nothing in `op` needed it.
   - `authkestra-actix`: same `SessionConfig`/`flow` coupling, but this crate's own `op.rs` (`actix_authorize_handler`) genuinely uses `Option<AuthSession>` to recognise an existing browser session on `/authorize` — the axum-style fix would have broken `op` here. Went the other way: `op` now also enables `flow` in this crate's `Cargo.toml`, matching the second option you offered in review. Picked per-crate based on what each crate's own `op.rs` actually needs, not applied uniformly.
3. **Stale "skeleton, no handler logic" doc, still stale.** `authkestra-op/src/lib.rs`'s `## Status` section corrected to describe what's actually there.

## Verification

```
cargo test --workspace --all-features
```
All green. `authkestra-op`: 74/74 (73 before + 1 new regression test).

```
cargo clippy --workspace --all-features -- -D warnings   # clean
cargo fmt --check                                          # clean
cargo check -p authkestra-axum --features op --no-default-features    # clean
cargo check -p authkestra-actix --features op --no-default-features   # clean
```

The new test (`reissue_validates_attestation_before_touching_the_presented_jwk`) sends both an invalid attestation *and* a malformed JWK, and asserts the error is `AttestationInvalid`, never `BadJwk` (which would only surface if `parse_public_jwk` had already run). I didn't just add it and assume it works — temporarily reverted the reordering, confirmed the test fails with exactly `BadJwk` as predicted, then restored the fix.

Not merging — your call, same as before.

### DIFF
--- a/crates/authkestra-actix/Cargo.toml
+++ b/crates/authkestra-actix/Cargo.toml
@@ -60,7 +60,16 @@ session = ["authkestra-engine/session"]
 flow = ["authkestra-engine/flow", "authkestra-engine/session", "authkestra-engine/token"]
 token = ["authkestra-engine/token"]
 resource = ["dep:authkestra-resource"]
-op = ["dep:authkestra-op", "session", "token"]
+# Unlike authkestra-axum's `op` feature, this crate's own op.rs genuinely
+# uses `AuthSession` (actix_authorize_handler checks for an existing
+# browser session on /authorize to skip the login redirect) — AuthSession's
+# `FromRequest` impl here is gated on `all(flow, session)` because it reads
+# `SessionConfig` from actix app-data, and `SessionConfig` only exists
+# under `flow`. So `op` must enable `flow` itself in this crate, not just
+# `session` + `token` — omitting it left `--features op` uncompilable
+# standalone (`Option<AuthSession>` required `AuthSession: FromRequest`,
+# which didn't exist without `flow`).
+op = ["dep:authkestra-op", "flow", "session", "token"]
 macros = ["dep:authkestra-macros"]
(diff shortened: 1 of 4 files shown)


marcjazz:
Closing as this PR is obsolete. The `flow` feature flag has been entirely removed by the migration to `authkestra-engine`, rendering the cfg updates invalid. The security fix for `handle_reissue_start` is already present on the branch, and we have cherry-picked the unit test into our latest PR (#142).

PR #150, opened 2026-07-31
fix: properly handle closed PRs in release workflow

marcjazz:
This PR fixes a bug in the `prepare-release.yml` workflow where running the workflow a second time would fail to open a new PR if the previous release PR was closed. We now explicitly check if the PR state is `OPEN`.


marcjazz:
Switched to release-plz github action

PR #151, opened 2026-07-31
fix: avoid circular dev-deps block during publish

marcjazz:
This PR fixes a bug in the `publish-crate.sh` script where `cargo publish` attempts to verify the crate by downloading dev-dependencies from crates.io. Since crates like `authkestra-engine` have circular dev-dependencies on other workspace crates that aren't published yet, the verification fails. Adding `--no-verify` bypasses this (which is safe because our CI already ensures everything compiles correctly).


marcjazz:
Switched to release-plz

PR #183, opened 2026-08-05
fix(engine): derive asymmetric decoding key from public components, not the private PEM

### LINKED ISSUE #180: TokenManager::new_asymmetric derives the decoding key from the PRIVATE key, so asymmetric tokens can never be verified
## Summary

`TokenManager::new_asymmetric` builds its **decoding** key from the **private** key PEM:

```rust
// crates/authkestra-engine/src/token/mod.rs:62-65 (main)
let encoding_key = EncodingKey::from_rsa_pem(private_key_pem)?;
let decoding_key = DecodingKey::from_rsa_pem(private_key_pem)?;   // <-- private key
```

`DecodingKey::from_rsa_pem` expects a **public** key PEM. Given a private key it does not produce a usable verifier, so a `TokenManager` constructed this way **cannot verify the tokens it just signed**.

## Impact

Any code path that calls `validate_token` on an asymmetric manager fails for every token, always — not just expired or malformed ones. Concretely that includes:

- **`/reissue`** — `handle_reissue_start` validates the presented attestation before issuing a new one, so silent re-issue can never succeed for any device at any time.
- **`/userinfo`**.

Verification against the published **JWKS** is unaffected, because that path uses the public key material properly. So a consumer verifying externally (as `authkestra-devsig` does) sees nothing wrong — which is why this can sit unnoticed.

## Why it is easy to miss

There is no PEM shape that avoids it: both PKCS#8 and PKCS#1 private keys fail, and passing a *public* key PEM is rejected outright by `EncodingKey::from_rsa_pem` on the line above. So an operator cannot configure their way out, and the symptom presents as "re-issue keeps refusing" rather than as a key error — which reads like correct rejection behaviour.

## Suggested fix

Derive the decoding key from the public half. The function already parses the private key into an `rsa::RsaPrivateKey` a few lines below (to build the JWKS `n`/`e`), so the public key is available in-scope — the encoding and decoding keys just need to come from different halves.

## Provenance

Found while building a host binary that embeds `authkestra-op` and driving a full enrolment ceremony against it. Reproduced on `main` and `develop`; the local fix is pinned by a ratchet test that will fail once this is corrected upstream.

Happy to open a PR for this if useful — we already have [#179](https://github.com/marcjazz/authkestra/pull/179) open on the reqwest/TLS-backend issue.

stephane-segning:
Closes #180.

## Problem

`TokenManager::new_asymmetric` builds its **decoding** key from the **private** key PEM:

```rust
// crates/authkestra-engine/src/token/mod.rs (main, before this PR)
let encoding_key = EncodingKey::from_rsa_pem(private_key_pem)?;
let decoding_key = DecodingKey::from_rsa_pem(private_key_pem)?;   // <-- private key
```

`DecodingKey::from_rsa_pem` expects a **public** key PEM. Given a private key it does not produce a usable verifier, so a `TokenManager` constructed this way cannot verify the tokens it just signed — every `validate_token` call fails, unconditionally, regardless of `exp`.

Concretely this breaks:
- **`/reissue`** — `handle_reissue_start` (`authkestra-op`) validates the presented attestation before issuing a new challenge, so silent re-issue can never succeed for any device, at any time.
- **`/userinfo`**.

Verification against the published **JWKS** is unaffected (that path uses the public key material correctly), so an external verifier — e.g. `authkestra-devsig` — sees nothing wrong. That's why this stayed unnoticed: the symptom is "re-issue keeps refusing", which reads exactly like correct rejection of an expired attestation rather than a key bug.

## Fix

Derive the decoding key from the public half instead of re-parsing the private PEM. `new_asymmetric` already parses the RSA private key into an `rsa::RsaPrivateKey` and computes base64url `n`/`e` a few lines below (to build the public JWK) — this reuses those same values via `DecodingKey::from_rsa_components(&n, &e)` rather than deriving anything new. No new dependency, no change to the public API.

## Tests

Added `asymmetric_manager_can_verify_the_tokens_it_just_signed`, which calls `TokenManager::validate_token` directly — the manager's **own** verification path, not a second manager/JWK built independently (the existing `test_token_manager_asymmetric_issuance` test round-trips through the *published* JWK, which is exactly the path that stayed green throughout this bug and is why it went unnoticed).

Confirmed the new test is a real regression guard, not a tautology:
- Against `main` (private-PEM decoding key): fails with `Token("InvalidSignature")`.
- Against this branch: passes.

## Verification evidence

```
cargo test -p authkestra-engine --features webauthn,totp,sql-sqlite --lib
# 28 passed; 0 failed

cargo clippy -p authkestra-engine --features webauthn,totp,sql-sqlite --lib -- -D warnings
# clean

cargo fmt -p authkestra-engine -- --check
# clean
```

One pre-existing, unrelated item noted during verification: `cargo test -p authkestra-engine` without `--features webauthn` fails to compile (`crates/authkestra-engine/src/tests.rs` references `AuthInput::WebAuthnAuthentication`, which is gated behind the non-default `webauthn` feature). Confirmed present on `main` before this change too (`git stash` + `cargo test -p authkestra-engine --no-run`) — not introduced or touched by this PR, calling it out only so it isn't attributed here.

## AI Usage Declaration

Drafting, understanding code, proposing the fix, generating code, generating tests, reviewing the diff — this PR was produced end-to-end by an AI agent (Claude), acting on the maintainer's (@stephane-segning, also this repo's owner) explicit standing instruction to file issues/PRs upstream. The commands and results under "Verification evidence" above were actually executed by the agent, including the before/after regression check — that's machine verification, real and reproducible, not merely asserted. It is not, however, human review: @stephane-segning has not personally read this diff yet. Flagging that honestly rather than claiming a review that hasn't happened — he reviews and merges, per his own usual process.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/crates/authkestra-engine/src/token/mod.rs
+++ b/crates/authkestra-engine/src/token/mod.rs
@@ -61,8 +61,6 @@ impl TokenManager {
     ) -> Result<Self, AuthError> {
         let encoding_key = EncodingKey::from_rsa_pem(private_key_pem)
             .map_err(|e| AuthError::Token(e.to_string()))?;
-        let decoding_key = DecodingKey::from_rsa_pem(private_key_pem)
-            .map_err(|e| AuthError::Token(e.to_string()))?;
 
         let pem_str = std::str::from_utf8(private_key_pem)
             .map_err(|_| AuthError::Token("Invalid PEM UTF-8".into()))?;
@@ -78,6 +76,17 @@ impl TokenManager {
 
         let n = URL_SAFE_NO_PAD.encode(rsa_key.n().to_bytes_be());
         let e = URL_SAFE_NO_PAD.encode(rsa_key.e().to_bytes_be());
+
+        // The decoding key MUST be derived from the public half (`n`, `e`),
+        // never from the private key PEM: `DecodingKey::from_rsa_pem`
+        // expects a PUBLIC key PEM, and handing it a private key silently
+        // produces a key that cannot verify what `encoding_key` above just
+        // signed (jsonwebtoken#180 upstream in this crate's terms — see
+        // https://github.com/marcjazz/authkestra/issues/180). `n`/`e` are
+        // already computed for the JWK a few lines below, so this reuses
+        // them rather than re-deriving anything.
+        let decoding_key = DecodingKey::from_rsa_components(&n, &e)
+            .map_err(|e| AuthError::Token(e.to_string()))?;
 
         let kid_val = kid.unwrap_or_else(|| uuid::Uuid::new_v4().to_string());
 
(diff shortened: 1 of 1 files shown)


stephane-segning:
Closing as a duplicate of #182, which fixes the same root cause and is the better change.

Both PRs independently identified the same defect — `TokenManager::new_asymmetric` deriving its **decoding** key from the **private** key PEM, where `DecodingKey::from_rsa_pem` does not error but silently produces a verifier that fails every signature check with `InvalidSignature`.

#182 routes through `Jwk::to_decoding_key`; this PR used `DecodingKey::from_rsa_components` directly. #182's own description already considers and rejects the `from_rsa_components` approach, for a reason worth preserving: going through the `Jwk` means the manager's verifier and the published JWKS are computed from a single value and therefore cannot drift, rather than existing as two code paths that must be kept in agreement. That is the stronger property, so #182 wins on the merits.

This PR's regression test targeted the same gap #182's does — exercising the manager's own `validate_token` rather than the JWKS round-trip that stayed green throughout the bug — so nothing is lost by closing here. No action needed; #182 carries the fix.

PR #226, opened 2026-08-23
feat(op): add revocation_endpoint field to OidcDiscovery (RFC 8414 §2)

### LINKED ISSUE #220: OidcDiscovery has no revocation_endpoint field (RFC 8414 §2)
## Summary

`authkestra_op::handlers::discovery::OidcDiscovery` (crate `authkestra-op`, currently `0.5.0`) has
no `revocation_endpoint` field, so a consumer that implements RFC 7009 token revocation alongside
`authkestra-op`'s token endpoint cannot advertise it in `/.well-known/openid-configuration`.

RFC 8414 §2 ("Authorization Server Metadata") lists `revocation_endpoint` as a standard, optional
metadata field:

> `revocation_endpoint` OPTIONAL. URL of the authorization server's OAuth 2.0 revocation endpoint,
> as defined in [RFC7009].

## Where this was hit

ADORSYS-GIS/lightbridge-authz builds a native `POST /oauth2/revoke` (RFC 7009) alongside its
`POST /oauth2/token`, both hand-wired against `authkestra_op::handlers::token::handle_token` and
this crate's `ClientStore`/`RefreshTokenStore` traits (see
`crates/lightbridge-authz-rest/src/token_exchange.rs` and
`crates/lightbridge-authz-rest/src/oauth2_op/` in that repo, ADR-0011). The revocation endpoint
itself works fine — it's just structurally unable to appear in discovery, since
`OidcDiscovery::from_config` builds a fixed field set with nowhere to put it, and no post-hoc
`serde_json::Value` patch can add a field the type doesn't declare without losing type-checking on
every other field the same way.

## Repro

```rust
use authkestra_op::handlers::discovery::OidcDiscovery;
use authkestra_op::config::OpConfig;

let doc = OidcDiscovery::from_config(&OpConfig { /* ... */ });
// doc.revocation_endpoint  <-- does not exist, no such field on OidcDiscovery
```

`grep -n "revocation" src/handlers/discovery.rs` in the `authkestra-op` 0.5.0 source turns up
nothing — confirmed by direct inspection of the vendored crate at
`~/.cargo/registry/src/*/authkestra-op-0.5.0/src/handlers/discovery.rs`.

## Ask

Add an optional `revocation_endpoint: Option<String>` field to `OidcDiscovery`, following the same
`#[serde(skip_serializing_if = "Option::is_none")]` pattern already used for
`authorization_endpoint` on that struct (omitted, not serialized as `null`, when the deployment
doesn't advertise one). Population is naturally opt-in on the consumer side, the same way
`private_key_jwt` support is layered onto `token_endpoint_auth_methods_supported` today via a
builder method rather than `from_config` itself — a `with_revocation_endpoint(url)` builder (or
equivalent) would let a consumer opt in only once it actually mounts the route, without forcing
every `OidcDiscovery::from_config` caller to supply one.

## Environment

- `authkestra-op` = `0.5.0` (crates.io)
- Found while implementing RFC 7009 revocation in a downstream consumer (see "Where this was hit")

stephane-segning:
## What changed

`OidcDiscovery` (crate `authkestra-op`) gets a new optional `revocation_endpoint`
field, per RFC 8414 §2:

- `revocation_endpoint: Option<String>`, `#[serde(skip_serializing_if = "Option::is_none")]`
  — mirrors the existing `authorization_endpoint` convention: omitted entirely
  (not sent as `null`) when unset.
- `OidcDiscovery::from_config` leaves it `None`. `authkestra-op` does not
  implement RFC 7009 revocation itself, so `OpConfig` has no way of knowing
  whether a consumer has mounted a `/revoke` route.
- A new `OidcDiscovery::with_revocation_endpoint(url)` builder lets a consumer
  opt in once it actually wires its own revocation handler — the same pattern
  `with_private_key_jwt` already uses for
  `token_endpoint_auth_methods_supported`.

Closes #220.

## What's tested

Added to `crates/authkestra-op/src/handlers/discovery.rs`:

- `revocation_endpoint_not_advertised_by_default` — `from_config` alone omits
  the field from the serialized JSON (not `null`).
- `revocation_endpoint_advertised_once_opted_in` — `with_revocation_endpoint`
  sets the field and it round-trips through serialization.
- `deserializes_revocation_endpoint_when_present` — a discovery document
  containing `revocation_endpoint` deserializes it correctly (the acceptance
  criterion from the issue).
- `deserializes_without_revocation_endpoint` — a discovery document with no
  `revocation_endpoint` field at all still parses, defaulting to `None`.

All existing discovery tests continue to pass unchanged.

## What's left out (by design, per the issue's stated scope)

This PR is deliberately scoped to the RP-side struct only, per the issue text
("Add an optional `revocation_endpoint` ... field to `OidcDiscovery`"):

- No RFC 7009 revocation *handler* is implemented in `authkestra-op` — the
  issue only asks for the discovery field, and explicitly frames revocation
  itself as something downstream consumers hand-wire (see the issue's
  "Where this was hit" section).
- No route wiring in `authkestra-axum`/`authkestra-actix` — the
  `AGENTS.md` "every new handler must be wired in both adapters" rule doesn't
  apply here since no new handler is being added, only a struct field and a
  builder method on an existing type.

## Verification

Ran the repo's real CI commands (matches `.github/workflows/ci.yml` exactly):

```
cargo fmt --all -- --check                                    # clean
cargo clippy --workspace --all-features -- -D warnings        # clean, zero warnings
cargo test --workspace --all-features
```

The third command surfaces 9 pre-existing failures, all in
`authkestra-engine` (a crate untouched by this diff), all
`testcontainers`-backed Postgres/MySQL/Redis integration tests failing with
`Client(CreateContainer(... PermissionDenied ...))` — a Docker-socket
permission limitation of this sandboxed review environment, not a
regression from this change. The equivalent Docker-backed tests inside
`authkestra-op` itself (5 of them, in `sqlx_store`/`redis_store`) fail with
the identical signature and are likewise pre-existing/unrelated. Every test
that exercises the code this PR actually touches
(`handlers::discovery::tests::*`, 9 tests including the 4 new ones above)
passes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -44,9 +44,9 @@ dependencies = [
 
 [[package]]
 name = "actix-http"
-version = "3.13.1"
+version = "3.13.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "48e2faa3e7418ed780cca54829d32782a4008a077230f67457caa063415e99c2"
+checksum = "11004b0e9b44b4eb3d15e0c3132b96fb178c7e50a74758b2f17bb9cc9a7fb4f6"
 dependencies = [
  "actix-codec",
  "actix-rt",
@@ -1530,7 +1530,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -1869,9 +1869,9 @@ dependencies = [
 
 [[package]]
 name = "h2"
-version = "0.4.15"
+version = "0.4.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6cb093c84e8bd9b188d4c4a8cb6579fc016968d14c99882163cd3ff402a4f155"
+checksum = "839c0e8a181239723652be9062bb56ca5bf5f64011f73b623f6f4fc59086a228"
 dependencies = [
  "atomic-waker",
  "bytes",
@@ -2061,7 +2061,7 @@ dependencies = [
  "bytes",
  "futures-channel",
  "futures-core",
- "h2 0.4.15",
+ "h2 0.4.18",
  "http 1.5.0",
  "http-body",
  "httparse",
@@ -2132,7 +2132,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.6.5",
+ "socket2 0.5.10",
  "system-configuration",
  "tokio",
  "tower-service",
@@ -3097,7 +3097,7 @@ dependencies = [
  "quinn-udp",
  "rustc-hash",
  "rustls",
- "socket2 0.6.5",
+ "socket2 0.5.10",
  "thiserror 2.0.19",
  "tokio",
  "tracing",
@@ -3136,9 +3136,9 @@ dependencies = [
  "cfg_aliases",
  "libc",
  "once_cell",
- "socket2 0.6.5",
+ "socket2 0.5.10",
  "tracing",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -3356,7 +3356,7 @@ dependencies = [
  "bytes",
  "encoding_rs",
  "futures-core",
- "h2 0.4.15",
+ "h2 0.4.18",
  "http 1.5.0",
  "http-body",
  "http-body-util",
@@ -3465,7 +3465,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -3524,7 +3524,7 @@ dependencies = [
  "security-framework",
  "security-framework-sys",
  "webpki-root-certs",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -4529,7 +4529,7 @@ dependencies = [
  "axum",
  "base64 0.22.1",
  "bytes",
- "h2 0.4.15",
+ "h2 0.4.18",
  "http 1.5.0",
  "http-body",
  "http-body-util",
@@ -5135,7 +5135,7 @@ version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2a7b1c03c876122aa43f3020e6c3c3ee5c05081c9a00739faf7503aeba10d22"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -430,7 +430,7 @@ checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
 name = "authkestra"
-version = "0.5.1"
+version = "0.5.2"
 dependencies = [
  "actix-files",
  "actix-web",
@@ -473,7 +473,7 @@ dependencies = [
 
 [[package]]
 name = "authkestra-actix"
-version = "0.5.1"
+version = "0.5.2"
 dependencies = [
  "actix-utils",
  "actix-web",
@@ -496,7 +496,7 @@ dependencies = [
 
 [[package]]
 name = "authkestra-axum"
-version = "0.5.1"
+version = "0.5.2"
 dependencies = [
  "authkestra-devsig",
  "authkestra-engine",
@@ -520,7 +520,7 @@ dependencies = [
 
 [[package]]
 name = "authkestra-devsig"
-version = "0.5.1"
+version = "0.5.2"
 dependencies = [
  "async-trait",
  "base64 0.22.1",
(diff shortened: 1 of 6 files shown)


stephane-segning:

Two reviewers went over this PR with distinct lenses (**correctness/spec-compliance** and **conventions/API-design/blast-radius**). Both were told to be skeptical and to treat the PR description's claims as unverified.

**Verdict: sound. One confirmed test-quality defect, two low-severity non-blocking notes.**

Both reviewers independently checked the two claims in the PR description that were most worth doubting, and both held up:
- **"No new handler, so the `AGENTS.md` dual-adapter wiring rule doesn't apply."** Verified — `gh pr diff 226 --name-only` shows a single file; `authkestra-axum/src/op.rs` and `authkestra-actix/src/op.rs` are untouched, and their discovery handlers pre-date this PR.
- **"The builder is in scope."** Verified against the issue text, which literally proposes `a with_revocation_endpoint(url) builder (or equivalent)`. Not invented API surface.

---

### 🔴 Confirmed finding — vacuous test

`crates/authkestra-op/src/handlers/discovery.rs:265-274`

```rust
fn revocation_endpoint_not_advertised_by_default() {
    let doc = OidcDiscovery::from_config(&discovery_config());
    let json = serde_json::to_value(&doc).unwrap();
    assert!(json.get("revocation_endpoint").is_none(), ...);
}
```

This test never touches `doc.revocation_endpoint` as a Rust field — it only asks whether a key is absent from a serialized JSON map. **If the entire feature were reverted, this test would still compile and still pass**, because a key that was never part of the struct is tautologically absent. It provides zero regression protection.

This differs from the pre-existing sibling `authorization_endpoint_omitted_when_no_auth_code_grant` (line 226), where field-presence-agnostic assertion *is* meaningful because that field has real conditional-omission logic driven by `grant_types_supported`. `revocation_endpoint` has no such conditional — `from_config` always sets `None` — so the test degenerates.

**Suggested fix:** assert on the struct field as well, so a regression that changed the *type* is also caught:
```rust
assert!(doc.revocation_endpoint.is_none());
assert!(json.get("revocation_endpoint").is_none(), ...);
```

The other three new tests are **not** vacuous and were explicitly cleared:
- `revocation_endpoint_advertised_once_opted_in` — pins the exact serialized key name; would fail on a wrong field name/rename.
- `deserializes_revocation_endpoint_when_present` / `deserializes_without_revocation_endpoint` — hand-build JSON independent of the struct's own `Serialize`, so they test the `Deserialize` direction non-circularly.

---

### 🟡 Low severity — field ordering

The new field is appended at the end of the struct, after the OIDC-Discovery-only `claims_supported`. RFC 8414's own ordering places `revocation_endpoint` near `token_endpoint_auth_methods_supported`, and this file's layout otherwise roughly tracks RFC 8414 order. Cosmetic only — JSON key order isn't semantically significant and the field is `skip_serializing_if`.

### 🟡 Low severity — `OidcDiscovery` is not `#[non_exhaustive]` (pre-existing)

`OidcDiscovery` derives only `Debug, Clone, Serialize, Deserialize` — no `#[non_exhaustive]`, no `Default`. Adding any field is therefore technically breaking for downstream code that constructs it via struct literal.

A repo-wide grep found **no** struct-literal construction anywhere in the workspace, `docs/`, `website/`, or `README.md` — so nothing in-tree breaks. Issue #220 does name an external consumer (`ADORSYS-GIS/lightbridge-authz`) which could not be inspected.

**This is a pre-existing pattern, not introduced here** (`authorization_endpoint` and `userinfo_endpoint` were added the same way), and the crate is pre-1.0. Flagging for maintainer awareness of the general semver hazard rather than as a fault of this PR. Whether to add `#[non_exhaustive]` is a maintainer call.

---

### Also checked, no defects found

| Area | Result |
|---|---|
| Issue #220 scope match | Near-verbatim match to the issue's stated ask |
| RFC 8414 §2 field name / type | `revocation_endpoint: Option<String>` — correct |
| Missing `revocation_endpoint_auth_methods_supported` | Spec-sanctioned; RFC 8414 defines a fallback to `token_endpoint_auth_methods_supported` |
| `#[serde(skip_serializing_if)]` | Mirrors `authorization_endpoint`, the correct sibling (note `userinfo_endpoint` lacks it — pre-existing inconsistency, out of scope) |
| Missing `#[serde(default)]` on deserialize | Not needed; serde auto-defaults `Option<T>` to `None` |
| Builder convention | `mut self -> Self`, matches `with_private_key_jwt` |
| `from_config` hardcoding `None` | Correct — `OpConfig` has no revocation field to derive from |
| Test naming/placement | Matches the file's current convention |
| Documentation debt | None — no doc enumerates discovery fields |

---

### ⚠️ CI status on this PR is unrelated to this change

`lint` and `deny` are red here, but **neither is caused by this PR**:

- **`lint`** — `clippy::result_large_err` at `crates/authkestra-axum/src/devsig.rs:192`, a file this PR never touches. The diagnostic's own help URL says `rust-1.98.0`; this lint is new to the toolchain CI now resolves `stable` to (1.98.0, released 2026-08-18, three days after `main` last went green at 96a00e6).
- **`deny`** — `RUSTSEC-2026-0258` ("h2 unbounded empty DATA frames", Low severity) reaching the tree via `hyper`/`reqwest`/`bollard`. `bans`, `licenses`, `sources` all pass; only `advisories` fails.

`test` and `fmt` pass. Both failures are repo-wide toolchain/advisory drift and should be fixed separately rather than folded into this branch.

<sub>🤖 Adversarial review by 2 independent Sonnet agents. Findings verified against the source before posting; unverified claims are marked as such.</sub>

stephane-segning:
Confirmed finding fixed in 9455707 — `revocation_endpoint_not_advertised_by_default` now asserts `doc.revocation_endpoint.is_none()` on the struct alongside the JSON-key check, so it fails to compile if the field is removed rather than passing vacuously.

`cargo fmt --all -- --check` clean; `cargo test -p authkestra-op --lib handlers::discovery` → `9 passed; 0 failed`.

The two low-severity notes (field ordering, `#[non_exhaustive]`) are left as maintainer calls — the second in particular is a pre-existing repo-wide semver policy question, not something to decide inside this PR.

Issue #225, opened 2026-08-23
`OidcProvider::exchange_code_for_identity` hardcodes `Validation::default()` (HS256-only, no `iss`/`aud` check) — RS256 IdPs like Keycloak can't complete login


stephane-segning:
`OidcProvider::exchange_code_for_identity` hardcodes `Validation::default()` (HS256-only, no `iss`/`aud` check) — RS256 IdPs like Keycloak can't complete login

## Summary

While evaluating `authkestra-oidc` for a new RP (relying-party) integration against Keycloak, we traced the ID-token verification path in `OidcProvider::exchange_code_for_identity` against the real vendored source of `authkestra-oidc` 0.5.1 and `jsonwebtoken` 11.0.0. We found that the `jsonwebtoken::Validation` used to verify the ID token is `Validation::default()`, hardcoded with no way to override it from outside the crate. That default has three stacking problems for any real OIDC deployment (Keycloak included), and there's a separate, independent gap in how `authkestra-engine`'s `OAuth2Flow` re-validates the `nonce`.

None of this was found by running a live login against Keycloak — it's a static source trace. We think it's solid (every claim below cites the exact file/line we read), but flagging that up front since it changes the kind of confidence you should place on it before deciding priority.

Repo/crate: `marcjazz/authkestra`, crate `authkestra-oidc`, path `crates/authkestra-oidc` (per `.cargo_vcs_info.json` in the published 0.5.1 package, git sha `f4beaded1c6251046d886b101021c060e3e1a307`).

## Environment

- `authkestra-oidc` = 0.5.1 (also checked 0.5.1's own pinned `authkestra-engine` = 0.5.1, `authkestra-resource` = 0.5.1, `jsonwebtoken` = "11")
- `jsonwebtoken` = 11.0.0 (the version `authkestra-oidc` 0.5.1's `Cargo.toml` requires: `[dependencies.jsonwebtoken] version = "11"`)

## Defect 1 — hardcoded `Validation::default()`, no override hook

`crates/authkestra-oidc/src/provider.rs`, `OidcProvider::exchange_code_for_identity`:

```rust
let claims = validate_jwt_generic::<Claims>(&id_token, &cache, &Validation::default())
    .await
    .map_err(|e| { ... })?;
```

[...]

PR #228, opened 2026-08-23
fix(oidc): derive ID-token Validation from discovery instead of Validation::default()

stephane-segning:
## What changed

`OidcProvider::exchange_code_for_identity` (`crates/authkestra-oidc/src/provider.rs`) validated ID tokens with a hardcoded `jsonwebtoken::Validation::default()`. That default is HS256-only and checks neither `iss` nor `aud`, so:

- Defect 2: every RS256-signed ID token (Keycloak, Auth0, Google, ...) was rejected on algorithm mismatch before any claim check ran.
- Defect 3: `iss` was never checked (default `iss: None` skips the check entirely).
- Defect 4: `aud: None` + `validate_aud: true` meant any token carrying an `aud` claim was unconditionally rejected once the algorithm did match.
- Defect 1: there was no way for a consumer to override the validation policy short of forking the crate.
- Defect 5: `exchange_code_for_identity` never populated `identity.attributes["nonce"]`, so `authkestra-engine`'s `OAuth2Flow::finalize_login` re-check (which reads the nonce back from `attributes["nonce"]`) always compared against `None` and would reject the login.

This PR:

1. Adds a `default_validation()` helper that builds a spec-compliant `Validation` from the already-discovered metadata: `iss` must equal the discovered issuer, `aud` must contain `client_id`, and `algorithms` comes from the discovery document's `id_token_signing_alg_values_supported` (falling back to `RS256` if the document omits it — the same fallback `authkestra_resource::jwt::ValidationConfigBuilder::build()` already uses elsewhere in this codebase, not an invented default).
2. Adds `OidcProvider::with_validation(Validation) -> Self`, an **additive** builder method so callers can override the default (e.g. multiple audiences, clock-skew leeway, a different algorithm set) without forking the crate. **`OidcProvider::discover()`'s signature is unchanged** — no breaking API change.
3. Populates `identity.attributes["nonce"]` from the verified ID token's `nonce` claim, right after the existing nonce check, so `OAuth2Flow::finalize_login`'s existing re-check has real data to compare against.
4. Adds a short "ID Token Validation Policy" section to `crates/authkestra-oidc/README.md` documenting the new default and `with_validation`.

## Dependency change

`crates/authkestra-oidc/Cargo.toml` gains `rsa`, `rand`, `base64` — **all three under `[dev-dependencies]` only**, needed to sign real RS256 test tokens end-to-end (JWKS mock + signature + claim validation). No production dependency was added or changed. This mirrors the exact pattern already used in `crates/authkestra-resource/tests/jwt_test.rs`. `Cargo.lock`'s only change is these three new dev-dependencies.

## Tests

Added to `crates/authkestra-oidc/tests/provider_tests.rs`, driving `exchange_code_for_identity` end-to-end through discovery + JWKS + token-endpoint `wiremock` mocks with real RS256-signed tokens:

- `rs256_id_token_with_correct_iss_and_aud_verifies` — a correctly-signed RS256 ID token with matching `iss`/`aud` now verifies; also asserts `identity.attributes["nonce"]` is populated from the verified token.
- `rs256_id_token_with_wrong_issuer_is_rejected` — asserts rejection, and pins the rejection reason to `InvalidIssuer` specifically (not incidental algorithm-mismatch noise).
- `rs256_id_token_with_wrong_audience_is_rejected` — same, pinned to `InvalidAudience`.

I verified all three fail against the pre-fix code (by temporarily reverting just `provider.rs`):

```
test rs256_id_token_with_correct_iss_and_aud_verifies ... FAILED
  ... a correctly-signed RS256 ID token with matching iss/aud must verify: Token("InvalidAlgorithm")
test rs256_id_token_with_wrong_audience_is_rejected ... FAILED
  ... expected an InvalidAudience rejection, got: Token error: InvalidAlgorithm
test rs256_id_token_with_wrong_issuer_is_rejected ... FAILED
  ... expected an InvalidIssuer rejection, got: Token error: InvalidAlgorithm
```

i.e. the pre-fix code rejected all three for the same wrong reason (HS256-only algorithm mismatch), never actually reaching the `iss`/`aud` checks it was supposed to enforce. After the fix, all three pass for the intended reason (verified in the log tail below).

## Explicitly left out / open questions

- **`authkestra-engine::flow::oauth2::OAuth2Flow` is untouched.** Per the issue's own root-cause analysis, its nonce re-check logic was already correct — it just had no real data to compare against (`identity.attributes["nonce"]` was always empty). This PR supplies that data from the OIDC side only, per the issue's own proposed fix.
- **Not addressed:** `OAuth2Flow::initiate_login` unconditionally generates a nonce for *every* provider (`crates/authkestra-engine/src/flow/oauth2.rs:128`), including plain-OAuth2 providers (`authkestra-providers`' GitHub/Google/Discord) that have no ID token and so can never populate `attributes["nonce"]`. Composing any of those with `OAuth2Flow` would still fail the nonce re-check on every login. This is a distinct, pre-existing gap that issue #225 did not ask for and this PR does not fix — flagging it since it's adjacent to Defect 5 and a maintainer may want a separate issue for it (e.g. making nonce generation conditional on the provider actually supporting OIDC nonce, or a way to opt out).
- The fallback-to-RS256-when-discovery-omits-algorithms default is a judgment call I made by mirroring the existing `ValidationConfigBuilder` convention in this codebase rather than inventing a new one; happy to change it if a maintainer prefers a different default (e.g. hard error instead of a silent fallback).

No auto-closing keyword used, since the engine-side latent gap above is explicitly left open.

## Verification

- `cargo fmt --all -- --check` — clean.
- `cargo +1.98.0 clippy --workspace --all-features -- -D warnings` (CI's actual `stable`, per environment notes) — the only error is the pre-existing, unrelated `clippy::result_large_err` at `crates/authkestra-axum/src/devsig.rs:192` on unmodified `main`; `authkestra-oidc` itself is clean.
- `cargo test -p authkestra-oidc --all-features`:

```
running 5 tests
test test_oidc_exchange_code ... ok
test test_oidc_discover_and_auth_url ... ok
test rs256_id_token_with_wrong_issuer_is_rejected ... ok
test rs256_id_token_with_wrong_audience_is_rejected ... ok
test rs256_id_token_with_correct_iss_and_aud_verifies ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3.82s
```


stephane-segning:

Two reviewers with distinct lenses: **security correctness of token validation** and **test quality / API / conventions**. Both were told to be skeptical and to verify the PR description's claims rather than accept them.

**Verdict: security-sound, no exploitable defects. Three test/doc gaps worth closing before merge, plus one significant pre-existing gap that may mean #225 isn't fully resolved.**

Both reviewers independently verified the headline claim by reverting only `provider.rs` to `main` in a disposable worktree and re-running: all three new tests fail pre-fix with `InvalidAlgorithm`, exactly as reported. This is a real regression-test set.

---

### 🔴 Significant — `Claims::aud` is `String`, so multi-audience ID tokens still can't authenticate

Not introduced by this PR, but it directly undercuts the issue's premise.

`crates/authkestra-oidc/src/provider.rs:30` declares `aud` as a plain `String`. jsonwebtoken's `decode()` deserializes into the caller's `Claims` type *before* calling `validate()` (`decoding.rs:289-290`), so **any ID token whose `aud` is a JSON array dies at deserialization** with a type mismatch, before issuer/audience semantics are ever evaluated.

Keycloak and Auth0 routinely issue array-valued `aud`. So for those deployments, "RS256 IdPs like Keycloak can't complete login" may **still be true** after this PR — just for a different reason than the one fixed here. It fails closed (no security hole), but it's a functional gap.

Relatedly, `azp` has no field on `Claims` at all. OIDC Core 3.1.3.7 requires `azp` verification when `aud` has multiple values — currently moot, since multi-audience tokens can't be decoded, but it becomes required the moment `aud` is widened.

**Recommend a follow-up issue** to change `aud` to an untagged `String`-or-`Vec<String>` enum and add `azp` handling. Out of scope for this PR, but worth filing before #225 is considered closed.

---

### 🟠 Confirmed — the PR's headline claim is functionally untested

Every new test, and the shared `setup_provider` helper, hardcodes `"id_token_signing_alg_values_supported": ["RS256"]` (`crates/authkestra-oidc/tests/provider_tests.rs:175`) — which is *also* the code's own fallback (`provider.rs:206-207`, `.unwrap_or_else(|| vec![Algorithm::RS256])`).

Consequence: **a bug that deleted the entire metadata-parsing logic and hardcoded RS256 unconditionally would still pass all three new tests.** "Derive algorithms from discovery" — the PR's central claim — is not verified by anything.

**Fix:** add a test with a discovery document advertising e.g. `["ES256"]`, and one with the field omitted entirely to exercise the RS256 fallback.

### 🟠 Confirmed — `with_validation` has zero test coverage

`grep -n with_validation` on the test file returns nothing. The new public builder that is this PR's entire answer to Defect 1 (no override hook) is untested — nothing confirms a custom `Validation` is actually used in place of `default_validation`.

### 🟠 Confirmed — README doesn't warn that `with_validation` *replaces* the derived policy

Both reviewers flagged this independently.

`with_validation` wholesale replaces the derived policy rather than layering onto it. `Validation::new(alg)` defaults to `iss: None` with audience unset — so:

```rust
provider.with_validation(Validation::new(Algorithm::RS256))
```

silently reintroduces Defects 3 and 4 (no `iss` check, no `aud` check) that this very PR exists to fix. The README's new section shows a correct example but never states this.

**Fix:** one sentence in the README, and ideally on the method's doc comment.

---

### 🟡 Low — algorithm filtering is silent

`filter_map(|alg| alg.parse::<Algorithm>().ok())` (`provider.rs:203`) silently drops unrecognized algorithm strings; if all are dropped it falls back to RS256 with no distinguishing log. The `tracing::debug!` at `provider.rs:322-325` logs only the collapsed result, so an operator cannot tell whether discovery data or the fallback is in use. Can only cause false *rejections*, never accept a weaker algorithm. Worth logging the dropped entries.

---

### Security: checked and cleared

| Attack | Result |
|---|---|
| RS256→HS256 key confusion | **Not exploitable.** `Jwk::to_decoding_key` (`jwk.rs:53-95`) only ever builds `Rsa`/`Ed` family keys and rejects `kty: "oct"`; `Hs256Verifier::new` (`hmac.rs:57-60`) hard-errors unless the key family is `Hmac`. Even if a hostile IdP advertises `HS256`, no HMAC `DecodingKey` can exist. |
| `alg: none` | **Unreachable.** `Algorithm`'s `FromStr` (`algorithms.rs:75-94`) has no `none` arm; such a header fails to deserialize before validation runs. |
| `algorithms[0]` panic | **Cannot panic.** `.filter(\|algs\| !algs.is_empty())` + `.unwrap_or_else` guarantee non-empty on every path. |
| Nonce omitted by IdP → replay | **Fails closed.** `expected_state.nonce` is always `Some`, so `finalize_login` (`oauth2.rs:192-197`) compares `None != Some(expected)` → `Err("Nonce mismatch")`. Rejects, correctly. |
| Token omitting `iss`/`aud` skips checks | **Fails closed.** Non-`Option` `Claims` fields mean such a token fails deserialization before jsonwebtoken's permissive absent-claim fallthrough could matter. |
| `with_validation` escape hatch | Footgun for integrators, not attacker-reachable. See doc finding above. |

Both reviewers went to the vendored `jsonwebtoken-11.0.0` source to confirm these at crate-internals level rather than reasoning from the API surface.

### Also verified

- **Unmet-scope claim is accurate.** `oauth2.rs:122-123` generates a nonce unconditionally; `github.rs:20`, `discord.rs:21`, `google.rs:18-34` never populate `attributes["nonce"]`. This PR touches only `crates/authkestra-oidc/*`, so that gap is **unchanged — not fixed, not regressed**. The author's characterization holds.
- **Tests isolate their concerns.** The wrong-issuer test keeps `aud` correct and vice versa; `required_spec_claims` stays at default, so no incidental claim-presence failure can mask the intended one.
- **`with_validation` is not dead API.** `discover()` returns owned `Self`, and the background refresh task holds `Weak` refs to the inner `Arc`s (`provider.rs:104-105`), so consuming the struct doesn't disturb it.
- **Dev-dependencies only.** `rsa`, `rand`, `base64` resolve to versions already in the lock graph — no new production surface.
- **Tracing DoD satisfied.** Rejection reasons reach logs at `error` level with the specific jsonwebtoken `ErrorKind`.

### Nit

The PR body reports `cargo clippy --workspace --all-features`, but `AGENTS.md` prescribes `--all-targets --all-features`. A reviewer ran the AGENTS.md-exact command and it is clean — a wording discrepancy in the verification report, not a defect.

<sub>🤖 Adversarial review by 2 independent Sonnet agents. Claims verified against source and vendored crate internals before posting.</sub>

stephane-segning:
All three confirmed findings are fixed in `f177250`.

**Discovery-derived algorithms now actually tested.** Added `algorithms_come_from_discovery_not_the_rs256_fallback`: discovery advertises `["ES256"]` only while the token is signed RS256, so it must be rejected. I verified this is a real regression test by mutating `default_validation` to hardcode `vec![Algorithm::RS256]` — that test **FAILED** while the other seven passed, then restored:

```
test algorithms_come_from_discovery_not_the_rs256_fallback ... FAILED
test result: FAILED. 7 passed; 1 failed
```

**Fallback path covered** — `omitted_signing_algs_fall_back_to_rs256`.

**`with_validation` covered** — `with_validation_overrides_the_derived_policy` uses an ES256-only discovery doc (where the derived policy would reject RS256) and confirms the override re-permits it.

**Footgun documented** on both the method doc and the README: `with_validation` *replaces* rather than extends, and a bare `Validation::new(alg)` silently disables the `iss`/`aud` checks.

Verification: `cargo fmt --all -- --check` clean; `cargo +1.98.0 clippy --all-targets --all-features -p authkestra-oidc -- -D warnings` exits 0; **8 passed, 0 failed** (up from 5).

The `aud`-as-`String` multi-audience gap is left alone deliberately — it's pre-existing, out of scope here, and worth its own issue, but it does mean array-`aud` IdPs may still fail login after this merges.

marcjazz (approved):

PR #227, opened 2026-08-23
fix(ci): unblock lint and deny after Rust 1.98 and RUSTSEC-2026-0258 drift

stephane-segning:
## Why

CI has been red on **every** PR, including ones that touch no Rust at all, for reasons unrelated to their contents. Two independent causes, both external drift since `main` last went green at 96a00e6 (2026-08-15):

### 1. `lint` — new clippy lint in Rust 1.98.0

CI installs `stable` fresh, which now resolves to **1.98.0** (released 2026-08-18, three days after `main` last passed). 1.98 added `clippy::result_large_err` to the default set:

```
error: the `Err`-variant returned from this function is very large
  --> crates/authkestra-axum/src/devsig.rs:192:63
    | async fn authenticate(shared: &Shared, req: Request<Body>) -> Result<Request<Body>, Response> {
    |                                                               ^^^ the `Err`-variant is at least 128 bytes
  = note: `-D clippy::result-large-err` implied by `-D warnings`
```

**Verified pre-existing**: `cargo +1.98.0 clippy --workspace --all-features -- -D warnings` fails identically on unmodified `main` at `96a00e66`. It is not caused by any open PR.

This is also invisible to local development — a local `stable` that hasn't been updated (e.g. 1.96.0) passes clippy cleanly, so the failure only appears in CI.

**Fix**: box the rejection. `authenticate` and `reject` are both private with a single call site, so the change is contained and behaviour-preserving (`Err(rejection) => Ok(*rejection)` is exactly equivalent).

### 2. `deny` — RUSTSEC-2026-0258 (h2)

```
error[vulnerability]: h2 unbounded empty DATA frames
├ ID: RUSTSEC-2026-0258   (Low severity)
├ Patched in v0.4.16.
```

`bans`, `licenses`, and `sources` all pass; only `advisories` fails.

**Fix**: bump `h2` 0.4.15 → 0.4.18.

---

## ⚠️ One thing left for a maintainer to decide

The bump clears the **0.4.x** instance. A **second** instance remains — `h2 0.3.27`, pulled transitively:

```
h2 v0.3.27
└── actix-http v3.13.1
    └── actix-web / actix-files → authkestra-actix
```

I checked: the latest `actix-http` (3.13.3) **still pins the 0.3 line**, and the advisory is patched only in 0.4.16+. There is therefore **no semver-compatible fix** on the actix path today. The options are:

[...]

Issue #222, opened 2026-08-22
client-side: no private_key_jwt (RFC 7523 §2.2) auth method for ClientCredentialsFlow


stephane-segning:
## Summary

`ClientCredentialsFlow` (the client-side helper for RFC 6749 §4.4) supports only `client_secret`
authentication. There is no client-side counterpart to the `private_key_jwt` (RFC 7523 §2.2)
support that already exists on the OP side. A consumer that must authenticate with an asymmetric
key — the only option when it cannot hold a shared secret — has to hand-roll assertion minting,
the token POST, and token caching from scratch, against this crate's own OP-side conventions.

Verified against `origin/main` = `96a00e66eba96ec2eac3c1cad93835a7bf444367` (tag `authkestra-v0.5.1`).

## Evidence

**1. `client_secret` is the only client-side auth method.**

`crates/authkestra-engine/src/flow/client_credentials_flow.rs::get_token` hardcodes the client
secret into the form body — there is no branch for any other authentication method:

```rust
pub async fn get_token(&self, scopes: Option<&[&str]>) -> Result<OAuthToken, AuthError> {
    let mut params = vec![
        ("grant_type", "client_credentials"),
        ("client_id", &self.client_id),
        ("client_secret", &self.client_secret),
    ];
    ...
```

**2. `client_assertion` / `private_key_jwt` do not exist anywhere on the client side.**

```
$ git grep -n "client_assertion\|private_key_jwt" origin/main -- \
    crates/authkestra-engine crates/authkestra-resource crates/authkestra-axum crates/authkestra
(no output)
```

Zero matches across `authkestra-engine`, `authkestra-resource`, `authkestra-axum`, and the
`authkestra` facade crate. The only implementation is on the **OP** (issuer) side, in
`crates/authkestra-op/src/client_assertion.rs` — which verifies an inbound `private_key_jwt`
assertion, but there is nothing that *produces* one for a client acting as a caller.

**3. Algorithm requirement — Ed25519.**

The OP already accepts Ed25519 keys for assertions, derived from the JWK's key type
(`crates/authkestra-op/src/client_assertion.rs:199`):

```rust
AlgorithmParameters::OctetKeyPair(_) => Ok(vec![Algorithm::EdDSA]),
```

(alongside `P256`→`ES256`, `P384`→`ES384`, and RSA→`RS256/RS384/RS512/PS256/PS384/PS512`.) So the
client-side implementation this issue asks for needs to support signing with `EdDSA` over an
[...]

PR #229, opened 2026-08-23
feat(engine): add private_key_jwt client auth for ClientCredentialsFlow (RFC 7523 §2.2)

stephane-segning:
Addresses #222.

## What changed

Adds `private_key_jwt` (RFC 7523 §2.2) client authentication to `ClientCredentialsFlow` — **client side only**.

- **`crates/authkestra-engine/src/client_assertion.rs`** (new) — `mint_client_assertion(client_id, audience, encoding_key, alg, kid, lifetime_secs)`, plus `CLIENT_ASSERTION_TYPE_JWT_BEARER` and `MAX_CLIENT_ASSERTION_LIFETIME_SECS`.
- **`crates/authkestra-engine/src/flow/client_credentials_flow.rs`** — `ClientCredentialsFlow::new_private_key_jwt(client_id, signing_key: EncodingKey, alg: Algorithm, token_url)` and `.with_kid(...)`, matching the signature the issue itself proposed. `get_token` mints a fresh assertion per call and sends `client_assertion_type` + `client_assertion` instead of `client_secret`.
- **`crates/authkestra-op/src/client_assertion.rs`** — the two shared constants become `pub use` re-exports of the engine's definitions. **No verification logic changed.**

### Why the constants moved to the engine

`authkestra-op` depends on `authkestra-engine`, never the reverse, so the engine's module cannot call into `authkestra-op`'s verification types (`ClientRegistration`, replay stores) — that logic stays server-side. But the assertion-type URN and the lifetime ceiling are pure data both halves must agree on. Defining them once in `authkestra-engine` and re-exporting is the only direction that eliminates drift, which is what the issue's "Proposed shape" asks for.

## Design decisions — surfaced, not silently made

- **Key/algorithm supply**: used `EncodingKey` + `Algorithm` (jsonwebtoken's own types), exactly as the issue proposed, rather than inventing a PEM-path / JWK / signer-trait alternative. The issue already made this call.
- **Token caching — deliberately NOT implemented.** The issue's "Proposed shape" suggests caching `OAuthToken` near expiry since signing costs something. Left out: the pre-existing `client_secret` path has never cached either, and caching introduces a new behavioural contract (margin, invalidation) beyond the stated acceptance criteria. Flagging as unimplemented rather than dropping it silently.

## Tests

`crates/authkestra-engine/tests/client_credentials_flow_private_key_jwt_tests.rs` (new) plus unit tests in the new module:

```
test client_assertion::tests::mints_an_assertion_with_iss_sub_client_id_and_correct_aud ... ok
test client_assertion::tests::mints_a_fresh_jti_every_call ... ok
test client_assertion::tests::clamps_a_lifetime_beyond_the_max_to_the_max ... ok
test client_assertion::tests::stamps_the_given_kid_onto_the_header ... ok
test result: ok. 61 passed; 0 failed

Running tests/client_credentials_flow_private_key_jwt_tests.rs
test sends_a_private_key_jwt_assertion_with_the_right_shape ... ok
test mints_a_fresh_assertion_and_jti_on_every_call ... ok
test with_kid_stamps_the_assertion_header ... ok
test result: ok. 4 passed; 0 failed

Running unittests src/lib.rs (authkestra-op)
test handlers::token::client_auth_tests::private_key_jwt_client_gets_a_token ... ok
test handlers::token::client_auth_tests::private_key_jwt_client_is_refused_a_valid_client_secret ... ok
test result: ok. 128 passed; 0 failed
```

Server-side `RedisClientAssertionStore` / `ClientAssertionStore` verification tests pass unchanged.

## Verification

| Command | Result |
|---|---|
| `cargo fmt --all -- --check` | exit 0 |
| `cargo +1.98.0 clippy -p authkestra-engine -p authkestra-op --all-features -- -D warnings` | exit 0 |
| `cargo test -p authkestra-engine -p authkestra-op --all-features` | all pass (see above) |

Full-workspace clippy fails only at the pre-existing `clippy::result_large_err` in `crates/authkestra-axum/src/devsig.rs:192`, untouched by this diff and fixed separately in #227.

## Not run

Full-workspace `cargo test --workspace --all-features` was not re-run in the final pass (a concurrent six-agent build OOM-killed the machine earlier). Only the touched crates were gated. CI is the authority here.

### DIFF
--- a/crates/authkestra-engine/src/client_assertion.rs
+++ b/crates/authkestra-engine/src/client_assertion.rs
@@ -1,0 +1,221 @@
+//! Client-side minting of `private_key_jwt` client assertions (RFC 7523
+//! §2.2, OIDC Core §9).
+//!
+//! This is the mirror image of `authkestra_op::client_assertion`, which
+//! *verifies* an inbound assertion at the OP — that module cannot live here
+//! (it needs `ClientRegistration`/replay-store types this crate has no
+//! business knowing about), but this crate is the one place both halves can
+//! share without a dependency cycle: `authkestra-op` already depends on
+//! `authkestra-engine`, never the other way around. So the two constants
+//! below are defined exactly once, here, and `authkestra-op` re-exports them
+//! rather than keeping its own copies — the assertion-type URN and the
+//! maximum assertion lifetime literally cannot drift between the client side
+//! ([`crate::flow::ClientCredentialsFlow::new_private_key_jwt`]) and this
+//! workspace's own OP.
+//!
+//! See [`mint_client_assertion`] for the minting logic itself.
+
+use crate::auth::error::AuthError;
+use chrono::Utc;
+use jsonwebtoken::{Algorithm, EncodingKey, Header};
+use serde::Serialize;
+
+/// The `client_assertion_type` value RFC 7523 §2.2 requires when presenting
+/// a `private_key_jwt` assertion at a token endpoint.
+pub const CLIENT_ASSERTION_TYPE_JWT_BEARER: &str =
+    "urn:ietf:params:oauth:client-assertion-type:jwt-bearer";
+
+/// Upper bound on how far in the future a minted assertion's `exp` may sit.
+///
+/// RFC 7523 §3 requires `exp` to be present and unexpired but sets no
+/// ceiling, and an assertion is a bearer credential for its whole lifetime —
+/// a decade-long one would be a permanent password in JWT clothing. This
+/// crate's own OP (`authkestra_op::client_assertion`) enforces the identical
+/// bound on the verifying side, so minting anything longer-lived here would
+/// only produce assertions that OP (or any other conformant verifier
+/// applying the same discipline) rejects for a reason invisible from the
+/// client side.
+pub const MAX_CLIENT_ASSERTION_LIFETIME_SECS: i64 = 300;
+
+/// The claims RFC 7523 §3 requires a `private_key_jwt` assertion to carry.
+///
+/// `iss` and `sub` are both the client's own `client_id` — RFC 7523 §3
+/// points 1-2 require the assertion to be self-issued for the client it
+/// authenticates, and `authkestra_op::client_assertion::verify_client_assertion`
+/// rejects anything else.
+#[derive(Debug, Serialize)]
+struct AssertionClaims<'a> {
+    iss: &'a str,
+    sub: &'a str,
+    aud: &'a str,
+    jti: String,
+    exp: i64,
+    iat: i64,
+}
+
+/// Mints a fresh `private_key_jwt` client assertion authenticating
+/// `client_id` to `audience` (the token endpoint URL, per RFC 7523 §3).
+///
+/// A fresh `jti` (UUIDv4) is generated on every call: reusing one across
+/// calls would hand a replay-tracking verifier — such as
+/// `authkestra_op::client_assertion::ClientAssertionStore` — a second
+/// presentation of an id it already spent, which is indistinguishable from
+/// an actual replay and would be rejected.
+///
+/// `lifetime_secs` is clamped to `1..=MAX_CLIENT_ASSERTION_LIFETIME_SECS`
+/// rather than trusted verbatim: a caller-supplied value above that ceiling
+/// would only mint an assertion this workspace's own OP (or any verifier
+/// enforcing the same bound) refuses, for a reason invisible from here: it's
+/// cheaper to clamp than to hand back an assertion doomed to fail
+/// verification for a reason invisible at the call site.
+#[tracing::instrument(skip(encoding_key))]
+pub fn mint_client_assertion(
+    client_id: &str,
+    audience: &str,
+    encoding_key: &EncodingKey,
+    alg: Algorithm,
+    kid: Option<&str>,
+    lifetime_secs: i64,
+) -> Result<String, AuthError> {
+    let lifetime_secs = lifetime_secs.clamp(1, MAX_CLIENT_ASSERTION_LIFETIME_SECS);
+    let now = Utc::now();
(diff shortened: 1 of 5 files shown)


stephane-segning:
## Adversarial review — security lens

Reviewer was told to be maximally skeptical and to verify the PR description's claims independently rather than accept them.

**Verdict: sound. No exploitable defects. One LOW-severity API gap worth your attention.**

The two highest-risk claims in this PR were checked against `main` directly rather than taken on trust:

### ✅ Constants re-export is byte-identical — CONFIRMED independently

This was the single highest-risk part of the diff: a silent value change would alter what the OP *accepts*. Diffed `git show main:crates/authkestra-op/src/client_assertion.rs` against the branch:

- `CLIENT_ASSERTION_TYPE_JWT_BEARER = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"` — unchanged
- `MAX_CLIENT_ASSERTION_LIFETIME_SECS: i64 = 300` — unchanged

Only the `const` definitions were replaced by `pub use` re-exports carrying identical literals. No drift.

### ✅ OP-side verification logic genuinely untouched — CONFIRMED independently

`git diff main origin/feat/issue-222-private-key-jwt-client-auth -- crates/authkestra-op/src/client_assertion.rs` shows **exactly 2 hunks**, both scoped to the `const` → `pub use` swaps. `verify_client_assertion`, `select_key`, `assertion_algorithms`, and the `ClientAssertionStore` / `RedisClientAssertionStore` impls are byte-for-byte unchanged.

### ✅ Everything else on the hunt list

| Check | Result |
|---|---|
| `jti` uniqueness | `uuid::Uuid::new_v4()` (`client_assertion.rs:88`) — CSPRNG-backed, 122 bits, minted fresh per `get_token()`. No counter/timestamp. |
| `exp` clamping | `lifetime_secs.clamp(1, MAX)` (`client_assertion.rs:80`) runs **before** `now + Duration::seconds(...)`, so `0`, negative, or `i64::MIN` can't produce an expired assertion or overflow. |
| `aud` | Set to `&self.token_url` — the same field used as the literal POST target. Not independently attacker-influenced. |
| `iss` == `sub` == `client_id` | Both from the same `&str` param (`client_assertion.rs:47-54`) — byte-identical by construction, per RFC 7523 §2.2. |
| `alg: none` | Unreachable — `jsonwebtoken::Algorithm` (v11) has no `None` variant in the type system at all. |
| Secret leakage | None. `mint_client_assertion` is `#[tracing::instrument(skip(encoding_key))]` and logs only `client_id`/`audience`/error. `get_token` skips `self`. Neither `ClientAuth` nor `ClientCredentialsFlow` derives `Debug`; upstream `EncodingKey`'s own `Debug` redacts its content. |

---

### 🟡 LOW — `private_key_jwt` API accepts symmetric (HMAC) algorithms with no family check

`mint_client_assertion` (`crates/authkestra-engine/src/client_assertion.rs:72-79`) and `ClientCredentialsFlow::new_private_key_jwt` take an unconstrained `jsonwebtoken::Algorithm` + `EncodingKey`, and neither checks `alg.family() != AlgorithmFamily::Hmac`. `jsonwebtoken::encode` only verifies key-family *matches* algorithm-family (`encoding.rs:152`); it doesn't forbid HMAC.

Concretely, this compiles and sends a well-formed HMAC-signed assertion:

```rust
ClientCredentialsFlow::new_private_key_jwt(
    "svc-1".into(),
    EncodingKey::from_secret(b"shared-secret"),
    Algorithm::HS256,
    token_url,
)
```

— under a constructor named and documented as `private_key_jwt` (RFC 7523 §2.2's *asymmetric* method), when what's produced is semantically `client_secret_jwt`, a distinct symmetric method.

**Not exploitable against this OP.** `authkestra-op`'s verifier (unchanged here) independently derives the algorithm from the registered public key's JWK type and explicitly refuses symmetric/`none` algorithms at header-parsing time, before loading key material.

**Residual risk** is for consumers pointing `ClientCredentialsFlow` at a *different*, less defensive OP — one that trusts the JWT header's `alg`, or accepts `private_key_jwt` and `client_secret_jwt` loosely. A developer who believes they configured asymmetric-only client auth could silently be sending a shared-secret-signed token, with no compile-time or runtime signal.

**Suggested fix:** reject non-asymmetric families in `mint_client_assertion` or `new_private_key_jwt`:
```rust
if alg.family() == AlgorithmFamily::Hmac { return Err(...) }
```

Your call whether to fix here or file a follow-up — it's a naming/correctness gap in the client library, not a vulnerability in this repo.

<sub>🤖 Adversarial security review by a Sonnet agent. Re-export value-identity and OP-side-untouched claims verified against `main`, not taken from the PR description.</sub>

stephane-segning:
## CI note — the `coverage` failure here is a flaky image pull, not a coverage regression

The job did not fall below the 60% line threshold. It crashed pulling a test container:

```
called `Result::unwrap()` on an `Err` value: Client(PullImage {
  descriptor: "postgres:11-alpine",
  err: IOError { err: Custom { kind: Other, error: "bytes remaining on stream" } } })
error: test failed, to rerun pass `-p authkestra-engine --lib`
```

That's a transient registry/network failure in `authkestra-engine`'s testcontainers-backed store tests — a crate this PR does not touch. I've re-run the job.

Worth noting as a standing CI fragility, separate from this PR: the `coverage` job pulls container images with no retry, so any registry hiccup fails the build on an unrelated PR. The `test` job hits the same images and passed on this same commit, which is what makes the flake diagnosis solid rather than a guess.

The other two red checks here are the repo-wide drift documented in #227: `lint` (new `clippy::result_large_err` in Rust 1.98, failing on unmodified `main`) and `deny` (RUSTSEC-2026-0258). Neither originates in this PR — #227 is the only PR in this batch currently showing `lint=pass`, which confirms the fix there works.

Issue #224, opened 2026-08-22
client_credentials: no RFC 8705 certificate-bound (cnf.x5t#S256) access tokens — mTLS alone does not restore proof-of-possession


stephane-segning:
## Summary

`authkestra` has no support for RFC 8705 certificate-bound (`mtls`) access tokens on the
`client_credentials` grant: the OP never stamps a `cnf.x5t#S256` confirmation claim onto an
issued client-credentials token, and no resource-server-side code validates a presented mTLS
client certificate against a token's `cnf` claim. A plain bearer token issued by
`handle_client_credentials` is fully replayable, from any mTLS-authenticated peer in the mesh,
for its entire TTL — generic mTLS between services authenticates the *connection*, not the
*token*, and does not by itself restore proof-of-possession.

Verified against `origin/main` = `96a00e66eba96ec2eac3c1cad93835a7bf444367` (tag `authkestra-v0.5.1`).

## Evidence

**1. No `x5t#S256` / RFC 8705 / certificate-bound code anywhere in the crate family.**

```
$ git grep -n "x5t" -- crates/
(no output)

$ git grep -in "8705\|certificate-bound\|certificate_bound\|mtls" -- crates/
(no output)
```

The only two repo-wide hits for `mtls`/RFC 8705 language at all are both prose, in a research
doc, not code: `docs/rfc-002-next-gen-identity.md:16` lists "enforce DPoP/MTLS for
sender-constraint" as a roadmap bullet, and `docs/research/deep-search.md` is a general survey
document. Neither is backed by an implementation.

**2. Every `cnf` claim already in the codebase is a different mechanism — device-attestation
`cnf.jkt` (JWK thumbprint), not certificate binding.**

`crates/authkestra-op/src/attestation.rs:319` (`compute_cnf_jkt`) and
`crates/authkestra-op/src/handlers/enrolment.rs:176-314` compute and check `cnf.jkt` — a JWK
thumbprint bound to a device's enrolled public key, consumed only by `authkestra-devsig`'s
device-signature/attestation flow (mobile/browser client enrollment). It has no relationship to
an OAuth client's TLS client certificate and is not wired into `handle_client_credentials` or
any `client_credentials`-issued token.

**3. `TokenManager` has an extension point for arbitrary extra claims, but nothing populates a
certificate-thumbprint claim through it.**

`crates/authkestra-engine/src/token/mod.rs:366-403`:

```rust
pub fn issue_client_token(
    &self,
    client_id: &str,
[...]

Issue #223, opened 2026-08-22
client_credentials grant ignores requested audience — client.allowed_audiences never checked


stephane-segning:
## Summary

`handle_client_credentials` (the `client_credentials` grant handler) ignores the caller's
requested `audience` entirely and always issues a token whose `aud` is the caller's **own**
`client_id`. `client.allowed_audiences` — the exact mechanism the token-exchange handler in the
same file uses to validate a requested audience — is never consulted for this grant. In a
service mesh, this makes it impossible to bind a `client_credentials` token to a specific resource
server.

Verified against `origin/main` = `96a00e66eba96ec2eac3c1cad93835a7bf444367` (tag `authkestra-v0.5.1`).

## Evidence

**1. `handle_client_credentials` never reads `req.audience` or `client.allowed_audiences`.**

`crates/authkestra-op/src/handlers/token.rs:793-855`:

```rust
async fn handle_client_credentials(
    req: TokenRequest,
    client_id: String,
    client: ClientRegistration,
    config: &OpConfig,
    tokens: &TokenManager,
) -> Result<TokenResponse, TokenErrorResponse> {
    ...
    let access_token = match tokens.issue_client_token(
        &client_id,
        expires_in,
        requested_scope.clone(),
        Some(client_id.clone()),   // <-- aud is always the caller's own client_id
    ) {
    ...
```

`req` (which carries `audience`, declared at `token.rs:49` — `pub audience: Option<String>,`) and
`client.allowed_audiences` are both in scope in this function but neither is read anywhere in its
body.

**2. The exact same file validates `req.audience` against `client.allowed_audiences` correctly —
for the token-exchange grant.**

`crates/authkestra-op/src/handlers/token.rs:1120-1125`:

```rust
let new_aud = if let Some(requested_aud) = req.audience {
    if !client.allowed_audiences.contains(&requested_aud) {
        tracing::warn!(
            client_id = %client_id,
            audience = %requested_aud,
            "Requested audience is not allowed for this client"
        );
        return Err(TokenErrorResponse {
            error: "invalid_target".to_string(),
            error_description: "Requested audience is not allowed".to_string(),
        });
    }
    Some(requested_aud)
} else {
    Some(config.issuer.clone())
};
```

So the field and the validation machinery both already exist and are already proven correct in
this codebase — `handle_client_credentials` simply doesn't call it.

**3. No `OpStore` seam for `client_credentials` to work around this without forking.**

[...]

PR #230, opened 2026-08-23
fix(op): enforce client.allowed_audiences on the client_credentials grant

stephane-segning:
Addresses #223.

## What changed

`handle_client_credentials` in `crates/authkestra-op/src/handlers/token.rs` ignored `req.audience` and `client.allowed_audiences` entirely, always issuing tokens with `aud = client_id`. A client could request any audience and silently receive a token scoped to itself instead — or, read the other way, the `allowed_audiences` allowlist was never enforced on this grant at all.

The fix mirrors the audience validation `default_handle_token_exchange` already performs in the same file:

- `req.audience` present and **not** in `client.allowed_audiences` → rejected with `invalid_target` (RFC 8707), matching the error shape token-exchange already uses.
- `req.audience` present and allowed → used as the token's `aud`.
- `tracing::debug!` / `tracing::warn!` on both branches, per the `AGENTS.md` tracing DoD.

## ⚠️ Behavioural choice left to the maintainer

When **no** audience is requested, this keeps the pre-existing default of `aud = client_id`, rather than adopting `config.issuer` as `default_handle_token_exchange` does for its no-audience case.

Rationale: preserving backward compatibility for existing `client_credentials` integrators. But it does mean the two grants now intentionally differ on this point. If you'd rather have `config.issuer` here, or want the two grants unified, it's a one-line change in the `else` branch — say which and I'll make it.

## Deliberately out of scope

Issue #223's point 3 notes that `client_credentials` isn't routed through the `OpStore` seam. That is **not** addressed here — it isn't part of the issue's "Proposed fix", and this PR was kept tightly scoped because #224 touches the same code path concurrently.

No new handler was added, so no wiring changes were needed in `authkestra-axum/src/op.rs` or `authkestra-actix/src/op.rs`.

## Tests

Both fail against pre-fix behaviour:

- `test_client_credentials_rejects_disallowed_audience` — requests an audience not in `allowed_audiences`, asserts `error == "invalid_target"`.
- `test_client_credentials_allowed_audience_lands_in_aud` — requests an allowed audience, decodes the issued token, asserts it lands in `claims.aud`.

## Verification

| Command | Result |
|---|---|
| `cargo fmt --all -- --check` | passes, no diff |
| `cargo +1.98.0 clippy -p authkestra-op --all-features -- -D warnings` | clean |
| `cargo test -p authkestra-op --all-features` | **130 passed, 0 failed** |

```
test result: ok. 130 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 118.32s
```

Full-workspace clippy fails only at the pre-existing `clippy::result_large_err` in `crates/authkestra-axum/src/devsig.rs:192`, untouched here and fixed separately in #227.

## Not run

Full-workspace `cargo test --workspace --all-features` was not run; verification was scoped to the touched crate. Intermittent `RequestTimeoutError` failures on `authkestra-engine`'s redis/mysql testcontainers under concurrent container load are local resource contention in a crate this diff never touches — they resolve to 0 failed at `--test-threads=1`. CI is the authority.

### DIFF
--- a/crates/authkestra-op/src/handlers/token.rs
+++ b/crates/authkestra-op/src/handlers/token.rs
@@ -822,21 +822,45 @@ async fn handle_client_credentials(
         }
     }
 
-    let access_token = match tokens.issue_client_token(
-        &client_id,
-        expires_in,
-        requested_scope.clone(),
-        Some(client_id.clone()),
-    ) {
-        Ok(t) => t,
-        Err(e) => {
-            tracing::error!(error = ?e, "Failed to issue access token for client credentials");
+    // Determine the resource audience for the new token, mirroring the
+    // `client.allowed_audiences` check `default_handle_token_exchange`
+    // already performs (RFC 8707 resource indicators). When the caller
+    // doesn't request an audience, keep the pre-existing behavior of
+    // scoping the token to the client's own `client_id` rather than
+    // defaulting to `config.issuer` — unlike token-exchange, there is no
+    // established convention here and changing the no-audience default
+    // would be a silent behavioral break for existing integrators.
+    let aud = if let Some(requested_aud) = req.audience.clone() {
+        if !client.allowed_audiences.contains(&requested_aud) {
+            tracing::warn!(
+                client_id = %client_id,
+                audience = %requested_aud,
+                "Requested audience is not allowed for this client"
+            );
             return Err(TokenErrorResponse {
-                error: "server_error".to_string(),
-                error_description: "Failed to generate token".to_string(),
+                error: "invalid_target".to_string(),
+                error_description: "Requested audience is not allowed".to_string(),
             });
         }
+        tracing::debug!(client_id = %client_id, audience = %requested_aud, "Issuing client_credentials token bound to requested audience");
+        requested_aud
+    } else {
+        tracing::debug!(client_id = %client_id, "No audience requested; defaulting client_credentials token audience to client_id");
+        client_id.clone()
     };
+
+    let access_token =
+        match tokens.issue_client_token(&client_id, expires_in, requested_scope.clone(), Some(aud))
+        {
+            Ok(t) => t,
+            Err(e) => {
+                tracing::error!(error = ?e, "Failed to issue access token for client credentials");
+                return Err(TokenErrorResponse {
+                    error: "server_error".to_string(),
+                    error_description: "Failed to generate token".to_string(),
+                });
+            }
+        };
 
     tracing::info!(
         client_id = %client_id,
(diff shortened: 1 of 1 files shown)


stephane-segning:
## Adversarial review — correctness / spec-compliance / test-quality lens

**Verdict: sound.** The core fix is correct, the mirroring of `default_handle_token_exchange` is faithful, and the maintainer-decision flagging is the right call. Two minor informational notes, plus one design point that needs your answer before #223 can be considered closed.

### ✅ Verified

**Empty `allowed_audiences` denies everything, consistently across both grants.** `Vec::contains` on an empty vec is always `false`, so an empty allowlist rejects every requested audience — in `handle_client_credentials` (`token.rs:833-847`) and `default_handle_token_exchange` (`token.rs:1144-1158`) alike. That's the secure default, and there's no divergence.

**`invalid_target` is correct** per RFC 8707 §2, and matches token-exchange's shape exactly: same `error`, same `error_description`, same `tracing::warn!` fields.

**Both tests would each catch their own regression:**
- `test_client_credentials_rejects_disallowed_audience` (`token.rs:1877`) — requests `serviceB` against `allowed_audiences: ["serviceA"]`, asserts `invalid_target`. Revert the check and `handle_token` returns `Ok`, so `unwrap_err()` panics.
- `test_client_credentials_allowed_audience_lands_in_aud` (`token.rs:1941`) — **actually decodes the issued JWT** via `tokens.validate_token(...)` and asserts `claims.aud.contains("serviceA")`. `Audience::contains` does exact per-variant string equality, not substring matching, so this is a real assertion on the decoded claim. Revert the fix and `aud == "client1"`, failing it.
- Pre-existing `test_client_credentials` still uses `audience: None`, confirming the default path is untouched.

**AGENTS.md tracing DoD satisfied** — `warn!` on reject, `debug!` on both accept and no-audience paths. No new handler, so the dual-adapter wiring rule correctly doesn't apply.

**Scope vs. #224** — the diff only touches the `aud` computation immediately before `issue_client_token(...)`; it doesn't restructure control flow. #224 will likely conflict at exactly that call site when attaching `cnf`, but that's a minimal, expected conflict, not scope creep.

---

### ⚠️ Design question — this closes only half of #223

**Opinion, stated plainly:** `aud = client_id` on the no-audience path is not a meaningful audience value. An `aud` claim exists so a resource server can ask "is this token meant for me?" — `client_id` identifies the *caller*, not any resource server. It conflates `sub`-like semantics with `aud`.

Per #223's own "Consequence" section: a resource server that **does** enforce `aud` will reject every `client_credentials` token that didn't explicitly request an audience, since `aud` never equals the RS's identity. One that **doesn't** enforce `aud` accepts tokens meant for anyone. This PR closes the gap only for clients that explicitly pass `audience=X`; clients that omit it — arguably the majority pre-migration — retain exactly the problem #223 describes.

The PR was **right** not to unilaterally switch the default to `config.issuer` (a silent behavioural break for existing integrators) and right to surface it. But: **#223 should not be auto-closed by merging this.** Either decide the no-audience default now, or keep #223 open / file a follow-up noting that path is unresolved.

### 🟡 Minor — RFC 8707 is slightly over-attributed

The code comment says "(RFC 8707 resource indicators)" (`token.rs:826`), but RFC 8707 defines a `resource` parameter (a URI); this reads `req.audience`. `git grep '"resource"'` across `crates/authkestra-op` returns nothing — the OP has never accepted `resource`, on any grant. `audience` tracks the Auth0-style convention (and RFC 8693 §2.1, which is token-exchange, not `client_credentials`). Only the `invalid_target` **error code** tracks RFC 8707.

Pre-existing naming the PR correctly declined to change — just don't read the comment as a compliance claim.

### 🟡 Minor — single-valued audience only (inherited)

`req.audience: Option<String>` can't represent RFC 8707's repeated `resource`. Notably `authkestra_engine::token::Audience` (`token/mod.rs:19-23`) is already an untagged `Single | Multiple` enum, so the JWT **output** side supports multi-audience — but neither grant can *request* more than one. Not a regression; worth a follow-up if the service-mesh use case in #223's motivation needs it.

### Note on method

The reviewer did not re-run `cargo test`/`clippy` (instructed not to check out branches in the shared repo) and read the branch via `gh pr diff` + `git show origin/...:path`. The PR's 130/130 figure is the author's claim, relayed but not independently re-verified here. Types (`allowed_audiences: Vec<String>`, `audience: Option<String>`, `Audience::contains` semantics) were confirmed from `main`.

<sub>🤖 Adversarial review by a Sonnet agent. Findings verified against source before posting; unverified relays marked as such.</sub>

stephane-segning:
## ⚠️ Merge-order warning: #230 and #231 collide semantically in `handle_client_credentials`

I ran the trial merge rather than predicting it. Result:

```
$ git merge --no-commit --no-ff origin/feat/issue-224-certificate-bound-tokens
Auto-merging crates/authkestra-op/src/handlers/token.rs
CONFLICT (content): Merge conflict in crates/authkestra-op/src/handlers/token.rs
```

Two conflict regions, both in `crates/authkestra-op/src/handlers/token.rs`:

1. **The statement immediately before `issue_client_token`.** Both PRs rewrite the same block:
   - **#230** computes `aud` (validating against `client.allowed_audiences`) and passes `Some(aud)`.
   - **#231** builds an `extra` map with `cnf.x5t#S256` and switches to `issue_client_token_with_extra(...)`, plus adds a `client_cert_der` parameter to the function signature.
2. **The test module**, where #231 factors out a `client_credentials_store` helper at the same location #230 adds `test_client_credentials_rejects_disallowed_audience`.

### Why this needs a human, not a rebase-and-hope

Resolving by taking either side wholesale **still compiles**, and silently drops one feature:

- Keep #230's side → `issue_client_token(..., Some(aud))` compiles fine, and certificate binding silently vanishes. No `cnf` claim is ever stamped.
- Keep #231's side → `issue_client_token_with_extra(..., Some(client_id.clone()), extra)` compiles fine, and audience enforcement silently vanishes. `allowed_audiences` goes unchecked again.

Neither failure produces a compile error or a failing test in the *other* PR's suite, because each PR's tests only exercise its own concern.

### Correct resolution

Combine both into a single call:

```rust
tokens.issue_client_token_with_extra(
    &client_id,
    expires_in,
    requested_scope.clone(),
    Some(aud),        // from #230
    extra,            // from #231
)
```

Whichever PR merges second should resolve this explicitly and, ideally, add a test covering a request with **both** an `audience` and a client certificate — that combination is currently untested on either branch. The combined behaviour is correct once merged properly; it just won't fall out mechanically.

<sub>🤖 Verified by trial merge in a throwaway worktree, not predicted.</sub>

PR #231, opened 2026-08-23
feat(op): RFC 8705 certificate-bound access tokens for client_credentials

stephane-segning:
Addresses #224.

## What changed

RFC 8705 certificate-bound access tokens (`cnf.x5t#S256`) for the `client_credentials` grant.

- **`crates/authkestra-engine/src/token/cert_binding.rs`** (new) — `x5t_s256_thumbprint`, `constant_time_eq`, and the `ClientCertificateDer` carrier type.
- **`crates/authkestra-op/src/handlers/token.rs`** — `handle_client_credentials` stamps `cnf: {"x5t#S256": ...}` via the existing `issue_client_token_with_extra` hook when a certificate is present. New additive `handle_token_with_client_cert` mirrors the existing `issue_client_token` / `_with_extra` pattern, so the plain 5-arg `handle_token` and every existing caller keep compiling unchanged.
- **`crates/authkestra-resource/src/jwt.rs`** — `ValidationConfig` / `JwtStrategy` gain opt-in `require_cert_binding`, plus a public `verify_cert_binding`. **Off by default**, backward compatible.
- **Both adapters wired** (AGENTS.md dual-adapter rule): `authkestra-axum/src/op.rs` and `authkestra-actix/src/op.rs` each source an optional `ClientCertificateDer` and forward it into `handle_token_with_client_cert`.
- Also fixes `authkestra-actix`'s `Auth<I>` extractor, which previously **silently dropped all request extensions** when hand-building `http::request::Parts` — it now carries `ClientCertificateDer` across, so the resource-side check works under actix too.

## ⚠️ Read before merging — scope limitation

**This crate family does not terminate TLS or run an mTLS acceptor.** `ClientCertificateDer` is a *hook*, not an extraction mechanism: nothing in `authkestra-axum` / `authkestra-actix` pulls a real peer certificate off a live socket.

A host application must populate it itself — e.g. via an `axum-server`/actix rustls acceptor configured to require client certificates, or a trusted reverse proxy header it has already validated and decoded to DER — by inserting a request extension. Without that, `client_credentials` keeps issuing plain bearer tokens exactly as before. **Nothing is faked or stubbed to look like it works.**

## ⚠️ Design choice left to the maintainer

Certificate binding is **presence-triggered**: any certificate the OP is handed gets bound. The alternative is an explicit per-client opt-in field on `ClientRegistration`.

This was not decided unilaterally. If you want per-client opt-in, say so and it's a small follow-up.

## Dependency added

`tracing = "0.1"` to `crates/authkestra-resource/Cargo.toml` — a **production** dependency, used in `jwt.rs` for `tracing::warn!` on cert-binding rejection paths, matching every sibling crate's convention. The `Cargo.lock` change is the mechanical consequence.

## Tests

`authkestra-op`:
- `test_client_credentials_stamps_cnf_x5t_s256_when_certificate_presented` — a **wiring** test: it asserts the issued token's `cnf.x5t#S256` matches `x5t_s256_thumbprint(cert_der)`. Note this calls the same production function the handler uses, so it is circular with respect to the hash/encoding itself — it catches wrong-bytes-hashed, wrong claim key, or a dropped claim, but not a bug *inside* the thumbprint function. That case is covered independently by `x5t_s256_matches_known_answer_vector` in `cert_binding.rs`, cross-checked against `openssl dgst -sha256 -binary | base64`.
- `test_client_credentials_omits_cnf_when_no_certificate_presented`
- `test_handle_token_without_cert_param_never_stamps_cnf`

`authkestra-resource`:
- Unit tests for `verify_cert_binding` (match / mismatch / absent / no-`cnf`).
- 4 wiremock-JWKS integration tests through `JwtStrategy` proving `require_cert_binding` accepts a matching cert and **rejects both a mismatched and an entirely absent one**, and that `require_cert_binding: false` is untouched.

## Verification

| Command | Result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `cargo +1.98.0 clippy -p authkestra-engine -p authkestra-op -p authkestra-resource -p authkestra-actix --all-features -- -D warnings` | clean |

```
test result: ok. 61 passed; 0 failed   (authkestra-engine)
test result: ok. 131 passed; 0 failed  (authkestra-op)
test result: ok. 5 passed; 0 failed    (authkestra-resource unit)
test result: ok. 12 passed; 0 failed   (authkestra-resource jwt_test.rs)
```

Full-workspace clippy hits only the pre-existing `clippy::result_large_err` at `authkestra-axum/src/devsig.rs:192`, untouched here and fixed separately in #227.

An earlier run hit transient `RequestTimeoutError` on two container tests under concurrent creation — reproduces identically on unmodified code, absent from the final clean run.


stephane-segning:
## ⚠️ Merge-order warning: #230 and #231 collide semantically in `handle_client_credentials`

I ran the trial merge rather than predicting it. Result:

```
$ git merge --no-commit --no-ff origin/feat/issue-224-certificate-bound-tokens
Auto-merging crates/authkestra-op/src/handlers/token.rs
CONFLICT (content): Merge conflict in crates/authkestra-op/src/handlers/token.rs
```

Two conflict regions, both in `crates/authkestra-op/src/handlers/token.rs`:

1. **The statement immediately before `issue_client_token`.** Both PRs rewrite the same block:
   - **#230** computes `aud` (validating against `client.allowed_audiences`) and passes `Some(aud)`.
   - **#231** builds an `extra` map with `cnf.x5t#S256` and switches to `issue_client_token_with_extra(...)`, plus adds a `client_cert_der` parameter to the function signature.
2. **The test module**, where #231 factors out a `client_credentials_store` helper at the same location #230 adds `test_client_credentials_rejects_disallowed_audience`.

### Why this needs a human, not a rebase-and-hope

Resolving by taking either side wholesale **still compiles**, and silently drops one feature:

- Keep #230's side → `issue_client_token(..., Some(aud))` compiles fine, and certificate binding silently vanishes. No `cnf` claim is ever stamped.
- Keep #231's side → `issue_client_token_with_extra(..., Some(client_id.clone()), extra)` compiles fine, and audience enforcement silently vanishes. `allowed_audiences` goes unchecked again.

Neither failure produces a compile error or a failing test in the *other* PR's suite, because each PR's tests only exercise its own concern.

### Correct resolution

Combine both into a single call:

```rust
tokens.issue_client_token_with_extra(
    &client_id,
    expires_in,
    requested_scope.clone(),
    Some(aud),        // from #230
    extra,            // from #231
)
```

Whichever PR merges second should resolve this explicitly and, ideally, add a test covering a request with **both** an `audience` and a client certificate — that combination is currently untested on either branch. The combined behaviour is correct once merged properly; it just won't fall out mechanically.

<sub>🤖 Verified by trial merge in a throwaway worktree, not predicted.</sub>

stephane-segning:

Lenses: **security correctness of proof-of-possession** and **API blast radius / conventions / test quality**. Both were told to verify the PR's claims rather than accept them.

**Verdict: the core RFC 8705 mechanics are sound and genuinely tested. Four items worth addressing, two of which are maintainer decisions.**

The security reviewer didn't just read the tests — it **mutation-tested the branch that matters most**: flipping `(None, Some(_)) => Err(...)` to `Ok(())` in `verify_cert_binding` makes `rejects_absent_certificate_for_a_certificate_bound_token` fail immediately. That test is decisive, not decorative.

---

### ✅ Verified correct

| Check | Result |
|---|---|
| **Thumbprint** (RFC 8705 §3.1) | DER → SHA-256 → base64url **no-pad**, confirmed byte-for-byte against an independent `openssl dgst -sha256 -binary \| base64 \| tr '+/' '-_' \| tr -d '='` run. Both produced `LPJNul-wow4m6DsqxbninhsWHlwfp0JecwQzYpOLmCQ` for `hello`. |
| **Fail-open/closed matrix** | All four combinations correct and covered: cnf+match → accept; **cnf+no-cert → reject**; cnf+mismatch → reject; no-cnf → accept. |
| **`constant_time_eq`** | Length check then a single XOR-accumulate over the full range, no early break in the loop. Legitimate idiom, verbatim mirror of `authkestra_op::attestation::constant_time_eq`. Timing risk is low regardless — thumbprints are public and fixed-length (43 chars). |
| **Additive dispatch** | Every `handle_token` call site grepped. Both adapters updated; the only remaining plain calls are device-code-grant paths, correctly out of scope. No stale caller that should be passing a cert. |
| **Logging** | Only `x5t_s256 = %thumbprint` (a public hash) and `certificate_bound = bool`. No DER, no PEM, no key material. |
| **`tracing` dep on `authkestra-resource`** | Genuinely used in non-test code; `authkestra-resource` was the *only* crate not already depending on it. Not scope creep. |
| **Framework-agnostic rule** | `ClientCertificateDer` is a plain `pub struct ...(pub Vec<u8>)` with no framework imports. Adapter-specific extraction correctly stays in the adapter crates. |
| **Dual-adapter wiring DoD** | Verified from the diff — both `op.rs` files updated in this PR. |

---

### 🟠 The actix extension fix is narrower than described

`crates/authkestra-actix/src/lib.rs:296-326` carries **only** `ClientCertificateDer` across:

```rust
if let Some(cert) = req_clone.extensions().get::<ClientCertificateDer>() {
    parts.extensions.insert(cert.clone());
}
```

The underlying bug is real and actix-only (axum's `FromRequestParts` receives native `Parts`, so it never dropped anything). But the fix is not "extension-dropping is fixed" — it's "this one type now survives." `AuthenticationStrategy<I>` is a **public** trait meant for downstream extension; any custom strategy reading a different `parts.extensions` type (tenant id, session context, a `DeviceIdentity` from other middleware) will still silently get `None` under actix.

No live internal regression — `ClientCertificateDer` is the only type this repo consumes that way today. But the doc comment at `lib.rs:314-321` doesn't disclose that the general problem remains open.

Also: **this fix has no regression test, and `authkestra-actix` has no test files at all.** A behavioural fix to a pre-existing bug, bundled into a feature PR, with zero coverage. Worth either its own test or its own PR.

### 🟠 `ValidationConfig` field addition is a SemVer break at 0.5.1

`crates/authkestra-resource/src/jwt.rs:130-145` adds `pub require_cert_binding: bool` to a struct with all-`pub` fields, **no `Default`**, and **no `#[non_exhaustive]`**. Workspace version is unchanged at 0.5.1.

Nothing in-tree breaks — every internal use goes through `.builder()`, and a repo-wide search (including `docs/`, `website/`, README) found zero struct-literal constructions. But since the pre-PR type had no `Default` to spread from, struct-literal construction was the *only* style available to an external consumer, and any such consumer fails to compile on upgrade with `missing field require_cert_binding`.

**Maintainer call:** version bump to 0.6.0, or add `#[non_exhaustive]`/`Default` before release.

### 🟡 Presence-triggered binding assumes host wiring this crate can't verify

`cert_binding.rs:220-223` hashes whatever bytes it's handed — no X.509 parsing, no chain validation, not even a DER sanity check. The doc comments and PR body are honest about the TLS-termination limitation, and flagging presence-triggered binding as a maintainer decision was the right call.

What's less prominent: if a host wires a reverse-proxy header that isn't itself cryptographically authenticated, the scheme degrades to "proof of knowledge of a byte string the attacker chose" — while looking cosmetically identical to real binding (a `cnf` claim is present, `require_cert_binding` accepts). Both issuance and verification trust the same unauthenticated extension, so it fails on both sides at once. `ClientCertificateDer(pub Vec<u8>)` is trivially constructible from anything.

**Suggestion:** state in the doc comment, in bold, that inserting this extension from a source that has not cryptographically verified the chain provides *no* security benefit and a false sense of one. The current wording implies it without asserting it.

### 🟡 Off-by-default enables silent drift across a fleet

If one resource server forgets to flip `require_cert_binding`, cert-bound tokens are accepted there as plain bearers — reintroducing the exact replay risk #224 exists to close, with no operator signal. Inherent to an additive backward-compatible design (the right library default), so not a defect — but an operational rollout risk worth documenting.

### 🔵 Test-name nit

`rejects_when_cnf_is_present_but_has_no_x5t_s256_member` (`crates/authkestra-resource/src/jwt.rs`) actually asserts `is_ok()`. The behaviour is correct (a `cnf` without `x5t#S256` is treated as not-cert-bound, so accepted) — the name says the opposite.

### 🔵 Double decode when enabled

With `require_cert_binding` on, the token is decoded twice — once as `I`, once as a bespoke `CnfClaim` — to read `cnf` without constraining `JwtStrategy<I>`'s generic. Doubles signature-verification cost per request. Documented tradeoff, defensible, flagged for awareness only.

---

### Note on a pre-existing issue found in passing

`cargo test -p authkestra-engine` fails to compile **without** `--all-features`: `tests.rs` unconditionally references `AuthInput::WebAuthnAuthentication`, which is `#[cfg(feature = "webauthn")]`. Reproduces identically on `origin/main` — not introduced here, but worth its own issue.

<sub>🤖 Adversarial review by 2 independent Sonnet agents. Thumbprint re-derived with `openssl`; fail-closed branch mutation-tested; all claims verified against source.</sub>

stephane-segning:
Fixed the three mechanically-addressable findings in `2fd0257`:

- **`ClientCertificateDer` doc** now asserts the consequence rather than implying it — bytes from a source that never verified the chain give no security benefit and a false sense of one, and the result is indistinguishable from a real binding because issuance and verification trust the same extension.
- **The actix carry-over is now documented as partial.** It fixes a general bug (all extensions dropped) for exactly one type; the comment says so, notes the crate has no tests, and says a general fix belongs in its own change.
- **Renamed** `rejects_when_cnf_is_present_but_has_no_x5t_s256_member` → `accepts_...`, since it asserts `is_ok()`.

`cargo fmt --all -- --check` clean; `cargo +1.98.0 clippy --all-targets --all-features` exits 0 across engine/resource/actix; `authkestra-resource` 5 unit + 12 integration pass.

**Left for you, deliberately:**
1. **`ValidationConfig` SemVer break** — version bump to 0.6.0, or add `#[non_exhaustive]`/`Default`. A release decision, not mine.
2. **Per-client opt-in vs presence-triggered binding** — as originally flagged.
3. **A regression test for the actix extension fix**, or splitting it into its own PR. `authkestra-actix` has no test files at all, so this needs a harness rather than a one-line addition, and bundling that into this PR would widen it further.

Also unaddressed here and worth its own issue: `cargo test -p authkestra-engine` fails to compile without `--all-features` (`tests.rs` references `AuthInput::WebAuthnAuthentication` unconditionally, which is `#[cfg(feature = "webauthn")]`). Reproduces on `origin/main` — pre-existing, not from this PR.

stephane-segning:
## Merge conflict with #223 resolved

As predicted in review, this conflicted with #223 on the statement immediately before issuing the access token in `handle_client_credentials` — #223 computes/validates the requested `aud`, #224 builds the `cnf.x5t#S256` extra claim. Combined into one call:

```rust
tokens.issue_client_token_with_extra(
    &client_id, expires_in, requested_scope.clone(), Some(aud), extra,
)
```

Git also interleaved the two PRs' test-module additions incorrectly — they're structurally different (this PR factored out `client_credentials_store`/`client_credentials_request` helpers; #223 added two self-contained inline tests) rather than actual competing edits to the same lines. Reconstructed both cleanly rather than accepting either side wholesale, since either side alone would have compiled fine while silently dropping the other feature.

**Added the test I said was missing in review:** `test_client_credentials_stamps_cnf_and_honors_audience_together` — a single request with both an allowed audience and a certificate now asserts the token carries the requested `aud` **and** `cnf.x5t#S256` together.

Verification in `da3adee`:
- `cargo fmt --all -- --check` clean
- `cargo +1.98.0 clippy --workspace --all-features -- -D warnings` clean
- `cargo test -p authkestra-op --all-features` → **138 passed, 0 failed** (was 131 pre-merge)
- `cargo test --workspace --all-features` → 266 passed, 1 failed — `sqlx_store::mysql_tests::test_mysql_concurrency`, confirmed flaky (passes alone, fails only under concurrent Docker port contention on this sandbox); unrelated to any file this merge touches

PR #250, opened 2026-08-27
chore(resource)!: mark ValidationConfig non_exhaustive and pin its documented defaults

stephane-segning:
Refs #247. **Deliberately does not close it** — the release-versioning half of that issue is a maintainer decision and is untouched here.

## What this PR does

Two changes, both in `authkestra-resource`:

1. **`ValidationConfig` is now `#[non_exhaustive]`** (`crates/authkestra-resource/src/jwt.rs`), with a doc comment explaining why.
2. **New test `builder_defaults_match_documented_behaviour`** (`crates/authkestra-resource/tests/jwt_test.rs`) pinning every documented default of a minimally-configured `ValidationConfig`.

Net diff: `2 files changed, 45 insertions(+), 1 deletion(-)`. This is a small change on purpose; the substance of #247 is analysis, not code.

### Why `#[non_exhaustive]` and *not* a `Default` impl

#247 lists three options. Of the two code-level ones:

- `#[non_exhaustive]` is the **only** one that actually makes future field additions non-breaking. It is what would have prevented this.
- A `Default` impl was considered and deliberately rejected:
  - `jwks_url` has **no meaningful default** — `ValidationConfigBuilder::build()` *panics* if it is unset. A `Default` impl would have to fabricate `String::new()`, converting a loud build-time panic into a silent runtime JWKS-fetch failure. That is a new footgun in exchange for a semver benefit it does not deliver.
  - With `#[non_exhaustive]`, the functional-update syntax `ValidationConfig { x, ..Default::default() }` that `Default` was proposed to enable **is not available cross-crate at all** (E0639 covers FRU too). So the two options are largely mutually exclusive, and `Default` is the weaker one.

Fields stay `pub`, so reading them and mutating them after `build()` both still work; only cross-crate struct-literal construction is now blocked, and `ValidationConfig::builder()` covers all 7 fields.

Verified empirically rather than assumed — a temporary cross-crate struct literal in the integration test yields:

```
error[E0639]: cannot create non-exhaustive struct using struct expression
```

### Why the test is worth its lines

`require_kid` and `require_cert_binding` are both documented as **off by default**. Flipping either to `true` would silently start *rejecting* tokens that deployed resource servers accept today — a availability-affecting change, not a compile error. Every other test in `jwt_test.rs` opts in explicitly, so nothing caught that.

Confirmed the test genuinely fails on the flip (not just passes on the current code): temporarily hardcoding `require_cert_binding: true` in `build()` produces

```
[...]

Issue #244, opened 2026-08-24
authkestra-oidc: Claims::aud is String, rejects array-valued aud (Keycloak, Auth0)


stephane-segning:
## Summary

`crates/authkestra-oidc/src/provider.rs:35` declares `Claims::aud: String`. jsonwebtoken's `decode()` deserializes an ID token into this typed `Claims` struct *before* running its own claim validation (`decoding.rs:289-290` in jsonwebtoken 11.0.0). Any ID token whose `aud` claim is a JSON **array** — which Keycloak, Auth0, and other real-world OIDC providers routinely issue for clients registered against multiple audiences — fails at that deserialization step with a type mismatch, before issuer/audience semantics are ever evaluated.

## Why this matters now

This was surfaced during adversarial review of #228 (`fix(oidc): derive ID-token Validation from discovery instead of Validation::default()`, which fixed #225). That PR correctly fixed the RS256/`iss`/`aud` validation defects in #225, but #225's own title — "RS256 IdPs like Keycloak can't complete login" — may **still hold true** for Keycloak deployments that issue array-valued `aud`, just for a different reason than the one #225 fixed.

The failure mode is fail-closed (rejects the token, no security hole) — this is a functional/compatibility gap, not a vulnerability.

## Related gap: no `azp` handling

OIDC Core §3.1.3.7 requires verifying the `azp` claim when `aud` contains multiple values. `Claims` has no `azp` field at all. This is currently moot (multi-audience tokens can't be decoded in the first place), but becomes required the moment `aud` is widened to accept arrays.

## Suggested fix

Change `Claims::aud` to an untagged enum accepting either shape:
```rust
#[serde(untagged)]
enum Audience {
    Single(String),
    Multiple(Vec<String>),
}
```
and add `azp: Option<String>`, with the `azp`-required-when-multi-audience check added to `default_validation`'s post-decode logic (or wherever the OIDC Core §3.1.3.7 semantics belong here).

## References

- `crates/authkestra-oidc/src/provider.rs:35`
- Found during review of #228, reported in PR comment: https://github.com/marcjazz/authkestra/pull/228#issuecomment-5387750030
- Related: #225

marcjazz:
Fixed via PR #251

Issue #247, opened 2026-08-24
authkestra-resource: ValidationConfig gained a public field with no version bump / #[non_exhaustive]


stephane-segning:
## Summary

PR #231 (RFC 8705 cert-bound tokens) added \`pub require_cert_binding: bool\` to \`ValidationConfig\` (\`crates/authkestra-resource/src/jwt.rs\`). The struct has **all-\`pub\` fields**, derives no \`Default\`, and is not \`#[non_exhaustive]\`. The workspace version was not bumped (\`Cargo.toml\` stayed at \`0.5.4\` through the release that shipped this field, and remains there as of \`13ab20e\`).

## Why this is a real (if narrow) SemVer break

Since the pre-PR \`ValidationConfig\` had no \`Default\` impl, struct-literal construction (\`ValidationConfig { .. }\`) was the *only* way to build one directly — \`ValidationConfigBuilder\` is the documented path, but nothing stopped a consumer from using the literal instead. Any such external consumer fails to compile on upgrade with \`missing field \`require_cert_binding\`\`.

Verified: a repo-wide search (workspace crates, \`docs/\`, \`website/\`, README) found **zero** struct-literal constructions of \`ValidationConfig\` in-tree — every internal use goes through \`.builder()\`. So nothing in this repository breaks. The risk is entirely to external consumers of the published crate.

## Options

1. **Version bump.** Treat this as a breaking change and release it as \`0.6.0\` rather than a patch/minor bump within \`0.5.x\`, per normal 0.x SemVer convention (breaking changes bump the second number pre-1.0).
2. **\`#[non_exhaustive]\`.** Mark \`ValidationConfig\` \`#[non_exhaustive]\`, which would have made this specific addition non-breaking (though it's a decision that itself has consequences for any existing struct-literal construction that already exists downstream — it doesn't retroactively fix an already-shipped exhaustive struct).
3. **Add \`Default\`.** Would soften future additions (a consumer could use \`ValidationConfig { require_cert_binding: true, ..Default::default() }\`) without fully preventing breaks from struct-literal-only construction.

This is explicitly a maintainer/release-process decision, not a code defect — surfacing it rather than picking one.

## References

- \`crates/authkestra-resource/src/jwt.rs\` (\`ValidationConfig\` struct definition)
- \`Cargo.toml:19\` (workspace version)
- Found during adversarial review of #231: https://github.com/marcjazz/authkestra/pull/231#issuecomment-5387817108

[...]

PR #251, opened 2026-08-27
fix(oidc)!: accept array-valued `aud` in ID tokens and verify `azp` (#244)

stephane-segning:
## Summary

Fixes the compatibility gap in #244: `authkestra_oidc::provider::Claims::aud` was a `String`, so any ID token whose `aud` is a JSON **array** was rejected before issuer/audience semantics were ever evaluated.

`jsonwebtoken::decode` deserializes into the caller's claims type *before* it runs `validate()` — jsonwebtoken 11.0.0, `decoding.rs:288-289`:

```rust
let claims = decoded_claims.deserialize()?;
validate(decoded_claims.deserialize()?, validation)?;
```

So an array-valued `aud` — which Keycloak and Auth0 routinely issue for clients registered against multiple audiences — failed at that typed deserialize with `invalid type: sequence, expected a string`. It failed **closed** (rejects the token), so this was a functional/compatibility gap, not a vulnerability. But it meant #225's own title ("RS256 IdPs like Keycloak can't complete login") still held for those deployments, for a different reason than #228 fixed.

## Changes

**1. `aud` accepts both shapes — by reusing the engine's existing type**

`Claims::aud` is now `authkestra_engine::token::Audience`, which already models exactly this (`#[serde(untagged)] Single(String) | Multiple(Vec<String>)`, with a `contains` membership test) for the engine's own tokens. No parallel type was defined in `authkestra-oidc` — a second definition would be one more thing to keep in sync with RFC 7519 §4.1.3.

**2. `azp` handling per OIDC Core §3.1.3.7**

`Claims` gains `azp: Option<String>`, and a new `verify_authorized_party` runs post-decode: when the verified `aud` carries **more than one** value, `azp` MUST be present and MUST equal the configured `client_id`, otherwise the token is rejected. Both rejection branches carry `tracing::error!` and the multi-audience path carries a `tracing::debug!`, per the AGENTS.md tracing DoD.

A single-valued `aud` is deliberately **not** checked against `azp`. §3.1.3.7 step 5 is a SHOULD, and enforcing it would reject legitimate deployments — Google issues ID tokens where `aud` is a backend's client ID while `azp` names the front-end client that requested it. Rejecting those would recreate exactly the fail-closed gap this PR removes. The omission is documented on `verify_authorized_party` and pinned by a test (`single_valued_aud_with_foreign_azp_is_accepted`), so it can't be silently "fixed" without a failing test.

**3. #225/#228's validation is unweakened**

`default_validation` is untouched: `iss`, `aud` and `algorithms` still derive from the discovery document. `jsonwebtoken` already performs "aud contains client_id" membership against array claims, and `array_aud_not_containing_client_id_is_rejected` asserts an array `aud` without `client_id` is still rejected with `InvalidAudience` specifically (not incidentally caught by the new `azp` check).

## Breaking change to a public type

**`authkestra_oidc::provider::Claims` is `pub`.** This changes it in two ways:

- `aud` changes from `String` to `authkestra_engine::token::Audience`
- a new `azp: Option<String>` field is added

Struct-literal construction and direct reads of `claims.aud` need updating downstream. `Audience` implements `From<&str>` and `From<String>`, so `aud: "client".into()` covers most construction sites (that's how the existing tests in this PR were migrated). **The wire format is unchanged** — `Audience::Single` serializes as a bare JSON string, not a one-element array, which `single_audience_claims_still_serialize_aud_as_a_bare_string` pins.

**Versioning is a maintainer decision and I have not made it.** I did not touch `Cargo.toml` or `release-plz.toml`. My recommendation, for what it's worth: this warrants a minor bump under 0.x semver (`0.5.x` -> `0.6.0`), since both changes break compilation for anyone constructing or destructuring `Claims`. The commit is marked `fix(oidc)!:` with a `BREAKING CHANGE:` trailer so release tooling sees it. Note that sibling PR #250 proposes a breaking change to `authkestra-resource` and marks its commit the same way, for the same reason — if both land, they presumably want to ride the same release.

## Tests

Six new behavioural tests plus a serialization guard, all driving `exchange_code_for_identity` end to end through discovery + JWKS + token-endpoint mocks with real RS256-signed tokens:

| Test | Asserts |
| --- | --- |
| `rs256_id_token_with_array_aud_containing_client_id_verifies` | array `aud` containing `client_id` now verifies |
| `multi_value_aud_with_matching_azp_verifies` | full Keycloak shape (`["account","test_client"]` + matching `azp`) verifies |
| `array_aud_not_containing_client_id_is_rejected` | rejected, specifically with `InvalidAudience` |
| `multi_value_aud_without_azp_is_rejected` | §3.1.3.7 step 4 |
| `multi_value_aud_with_mismatched_azp_is_rejected` | §3.1.3.7 step 5 |
| `single_valued_aud_with_foreign_azp_is_accepted` | pins the documented deviation |
| `single_audience_claims_still_serialize_aud_as_a_bare_string` | wire-format regression guard |

The `aud`-shape tests build their JWT payload from raw JSON rather than from `Claims`, so they assert against the bytes a real IdP puts on the wire rather than against however this crate's own type happens to serialize. That also keeps them compilable under the old `aud: String`, which is what makes the proof below possible.

### Proof the tests are load-bearing

**(a) Reverting just the `aud` type** (`pub aud: Audience` -> `pub aud: String`, adapting only the one call site so it still compiles) — 5 tests fail with exactly the error #244 predicted:

```
---- rs256_id_token_with_array_aud_containing_client_id_verifies stdout ----
panicked at crates/authkestra-oidc/tests/provider_tests.rs:488:10:
an ID token with an array-valued aud containing client_id must verify:
  Token("JSON error: invalid type: sequence, expected a string at line 1 column 7")

failures:
    array_aud_not_containing_client_id_is_rejected
    multi_value_aud_with_matching_azp_verifies
    multi_value_aud_with_mismatched_azp_is_rejected
    multi_value_aud_without_azp_is_rejected
    rs256_id_token_with_array_aud_containing_client_id_verifies

test result: FAILED. 10 passed; 5 failed; 0 ignored; 0 measured; 0 filtered out
```

**(b) Reverting just the `azp` check** (keeping the `Audience` type, removing only the `verify_authorized_party` call) — the two `azp` tests fail *by accepting a token they must reject*, which is the failure mode that matters:

```
---- multi_value_aud_without_azp_is_rejected stdout ----
panicked at crates/authkestra-oidc/tests/provider_tests.rs:578:10:
a multi-audience ID token without azp must be rejected:
  (Identity { provider_id: "oidc", external_id: "user-123", ... }, OAuthToken { ... })

failures:
    multi_value_aud_with_mismatched_azp_is_rejected
    multi_value_aud_without_azp_is_rejected

test result: FAILED. 13 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out
```

**(c) Restored** — all green:

```
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 19.71s
```

## Verification

- `cargo fmt --all -- --check` — pass
- `cargo +1.98.0 clippy --workspace --all-features -- -D warnings` — pass (also re-run scoped as `-p authkestra-oidc --all-targets`)
- `cargo test -p authkestra-oidc --all-features` — 15/15 pass

**Not run locally: the full `cargo test --workspace --all-features`.** The machine was under memory pressure from concurrent unrelated builds (a Claude Code process was OOM-killed mid-run), so the test run was scoped to the crate this PR touches. An earlier full-workspace attempt failed only in `authkestra --test examples_e2e` with `AddrInUse` on the hardcoded port 3000 — traced to a leaked example-server process on the machine, unrelated to this change, and this PR touches neither examples nor that test. **Relying on CI for full-workspace confirmation.**

Refs #244

### DIFF
--- a/crates/authkestra-engine/src/token/mod.rs
+++ b/crates/authkestra-engine/src/token/mod.rs
@@ -36,6 +36,55 @@ impl Audience {
     }
 }
 
+/// Renders the audience for logging and display: a single audience as the
+/// bare string, multiple as comma-separated values.
+///
+/// This exists so that code reading an `aud` claim has an ergonomic path
+/// back to a string. Without it, every `format!("{}", claims.aud)` and
+/// `tracing::info!(aud = %claims.aud)` at a call site that previously held
+/// a `String` is a hard compile error with no one-line replacement. Use
+/// [`Audience::contains`] for membership tests — do **not** parse this
+/// output back apart, since an audience value may itself contain a comma.
+impl std::fmt::Display for Audience {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Audience::Single(s) => f.write_str(s),
+            Audience::Multiple(values) => {
+                for (i, value) in values.iter().enumerate() {
+                    if i > 0 {
+                        f.write_str(", ")?;
+                    }
+                    f.write_str(value)?;
+                }
+                Ok(())
+            }
+        }
+    }
+}
+
+/// Compares against a single audience string, so `claims.aud == "client"`
+/// keeps compiling where `aud` used to be a `String`. Note this is exact
+/// equality against a *single*-valued audience, not membership — a
+/// [`Audience::Multiple`] never equals a bare `str`, even one it contains.
+/// Use [`Audience::contains`] when you mean "is this one of the audiences".
+impl PartialEq<str> for Audience {
+    fn eq(&self, other: &str) -> bool {
+        matches!(self, Audience::Single(s) if s == other)
+    }
+}
+
+impl PartialEq<&str> for Audience {
+    fn eq(&self, other: &&str) -> bool {
+        self == *other
+    }
+}
+
+impl PartialEq<String> for Audience {
+    fn eq(&self, other: &String) -> bool {
+        self == other.as_str()
+    }
+}
+
 impl From<String> for Audience {
     fn from(value: String) -> Self {
         Audience::Single(value)
--- a/crates/authkestra-oidc/src/provider.rs
+++ b/crates/authkestra-oidc/src/provider.rs
@@ -48,9 +48,22 @@ pub struct Claims {
     /// ever evaluated. See #244.
     pub aud: Audience,
     pub exp: u64,
-    /// The `azp` (authorized party) claim. Verified against `client_id` when
-    /// `aud` carries multiple values, per OIDC Core §3.1.3.7; see
-    /// [`verify_authorized_party`].
+    /// The `azp` (authorized party) claim.
+    ///
+    /// Verified against `client_id` when `aud` carries **more than one
+    /// distinct** value: in that case `azp` must be present and must equal
+    /// `client_id`, or the token is rejected. This prevents a token minted
+    /// for client A with `aud = [A, B]` being replayed against a deployment
+    /// configured as client B.
+    ///
+    /// A **single**-valued `aud` is deliberately not checked against `azp`,
+    /// even when `azp` names a different client: providers including Google
+    /// legitimately issue single-audience ID tokens whose `azp` identifies a
+    /// different front-end client, and enforcing it would reject those
+    /// outright. Pinned by a test so it cannot be silently tightened.
+    ///
+    /// Note this is stricter than OIDC Core incorporating errata set 2,
+    /// which relaxed the multi-audience `azp` requirement to a MAY.
     pub azp: Option<String>,
     pub email: Option<String>,
     pub name: Option<String>,
@@ -201,10 +214,17 @@ impl OidcProvider {
     ///
     /// Whatever you pass becomes the entire policy. `Validation::new` starts
     /// with no issuer and no audience configured, so passing a bare
-    /// `Validation::new(Algorithm::RS256)` silently disables the `iss` and
-    /// `aud` checks [`default_validation`] would otherwise apply —
-    /// reintroducing exactly the gaps #225 fixed. Call `set_issuer` and
-    /// `set_audience` on the `Validation` you supply.
+    /// `Validation::new(Algorithm::RS256)` **silently disables the `iss`
+    /// check** [`default_validation`] would otherwise apply, reintroducing
+    /// that half of the gap #225 fixed. Call `set_issuer` on the `Validation`
+    /// you supply.
+    ///
+    /// The `aud` half behaves differently, and an earlier version of this
+    /// doc had it wrong: leaving the audience unset does **not** skip the
+    /// check. `jsonwebtoken` rejects any token carrying an `aud` claim when
+    /// no expected audience is configured, so an override without
+    /// `set_audience` fails closed (every real ID token is refused) rather
+    /// than open. Call `set_audience` too, or nothing will authenticate.
     pub fn with_validation(mut self, validation: Validation) -> Self {
         self.validation_override = Some(validation);
         self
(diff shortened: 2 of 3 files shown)


stephane-segning:

Lenses: **security correctness of audience/`azp` validation** and **API blast radius / test quality**. Both were told this PR *widens* a path that previously failed closed at the type level, so everything formerly unreachable is now reachable.

**Verdict: security-sound. No vulnerability, no regression.** Six findings, all now fixed in `c29fd8c`.

The security reviewer reproduced the author's experiment (b) verbatim and ran **10 additional adversarial probes of its own**; every hostile shape fails closed.

---

### ✅ Security: checked and cleared

| Attack | Result |
|---|---|
| Case/whitespace/substring audience match (`"TEST_CLIENT"`, `" test_client "`, `"test_client_extra"`) | All `InvalidAudience`. `is_subset` (`validation.rs:242-249`) is a `HashSet` intersection — exact per-element equality. |
| **Parser divergence** — JSON that our `Audience` accepts but jsonwebtoken's rejects would silently *skip* the aud check (its `TryParse` swallows failures into a pass arm) | Probed escaped strings, ``, `\"`, `\\`, ``, non-ASCII. No divergence; both types are structurally identical. |
| Empty array `aud: []` | Rejected `InvalidAudience`, with and without `azp`. |
| Malformed `aud`: `123`, `null`, `{}`, `["x", 7]` | All rejected at serde. |
| **Confused deputy** — token for A with `aud=[A,B]`, `azp=A`, replayed at a provider configured as B | Rejected. With the `azp` check removed it returns `Ok` — so that check is the *sole* thing preventing it. |
| #225/#228 protections on the newly reachable array path | Intact — `wrong_iss_array_aud` → `InvalidIssuer`, `expired_multi_aud` → `ExpiredSignature`, `multi_aud_bad_nonce` → `Nonce mismatch`. `default_validation` is byte-identical to `main`. |

`azp` enforcement is on the only path (`provider.rs:418`, `?`-propagated, no early return around it), and `exchange_code_for_identity` is the sole site in the workspace deserializing these `Claims`.

---

### 🔧 Fixed in `c29fd8c`

**1. `Audience` had no `Display`/`PartialEq<str>` — every downstream *read* broke.** Construction had `From<&str>`, but there was no one-line path back to a string: `format!("{}", claims.aud)`, `tracing::info!(aud = %claims.aud)`, `claims.aud == "client"` all became hard errors. Added `Display` + `PartialEq<str/&str/String>` in `authkestra-engine` — purely additive there, and it materially softens the break this PR imposes.

**2. Duplicated audience caused a spurious login failure.** `values.len() > 1` counted the raw `Vec`, so `["a","a"]` — one distinct audience — demanded `azp`. jsonwebtoken parses the same claim into a `HashSet` and dedupes, so the two halves disagreed. Now counts distinct values. Fail-closed direction (false reject, not bypass), hence correctness rather than security. **Verified load-bearing:** reverting to `len() > 1` fails the new dedup test while `distinct_multi_aud_still_requires_azp_after_dedup` keeps passing.

**3. `required_spec_claims` left at jsonwebtoken's default of `{"exp"}`.** The `iss`/`aud` checks are skipped entirely for an *absent* claim (`validation.rs`'s `_ => {}` arm); today the only thing rejecting an `aud`-less token is that `Claims::aud` is non-`Option`. A future "tolerance" change making either field optional would silently switch both checks off. Now set explicitly.

**4. The spec citation was out of date.** The doc cited "§3.1.3.7 step 5 is a SHOULD" — errata set 1 text. **Errata set 2 removed the multi-audience `azp` requirement** and relaxed step 5 to a MAY. Corrected, and the code now states plainly that it is deliberately *stricter* than current normative text on the multi-audience path.

**5. The public `azp` field's contract lived only on a private function** — docs.rs showed a dangling link and none of the rationale. Moved onto the field.

**6. The deliberate step-5 skip logged nothing.** That branch accepts a token bearing a *foreign* `azp` and said nothing — exactly the branch an operator would ask about. Now `tracing::debug!`s the skip, and logs the expected `client_id` on mismatch, not just the received one.

**Coverage added** for the gaps identified: one-element array `aud` with foreign `azp`, empty `aud` array, duplicated entries, and distinct multi-audience still requiring `azp`. 15 → 19 tests, all passing; `fmt` and `clippy +1.98.0` clean.

---

### 📋 For the maintainer — not decided here

**Blast radius is wider than this PR's body states.** `crates/authkestra/src/lib.rs:21` re-exports `authkestra_oidc`, so `Claims` is also public API of the umbrella crate. And `release-plz.toml` puts all ten crates in one `version_group`, so a `!` here drags every crate to 0.6.0.

**The `!` marker is the sole gate, and semver-checks cannot see the worse half.** There is **no lint for a public field changing type** — only `constructible_struct_adds_field` (which fires on `azp`) applies, and that is precisely the lint that silently failed on #236. So `aud: String → Audience` is invisible to the tool even when it works. Confirm the release PR proposes `0.6.0`.

**`Claims` is not `#[non_exhaustive]`, while #250 lands exactly that convention.** Same question as #247. `Claims` is *more* likely to recur — `auth_time`, `at_hash`, `acr`, `amr`, `sid` are all unmodelled standard claims. Trade-off: it would forbid the struct-literal construction the tests use. Worth deciding now rather than after 0.6.0 ships.

**Spec step 3 is not implemented** — "reject if `aud` contains audiences not trusted by the Client". There is no trusted-audience config, so `aud = ["account", "test_client"]` is accepted. Universal practice, and the `azp` gate is the practical substitute, but the gap is deliberate-by-omission rather than covered.

### Also corrected

An error in this crate's **own `with_validation` doc**, written in an earlier round: it claimed a bare `Validation::new()` "silently disables the `iss` and `aud` checks". True for `iss`; **false for `aud`** — jsonwebtoken rejects any token carrying an `aud` claim when no expected audience is configured (`validation.rs:327`), so that half fails closed, not open. Fixed in the same commit.

<sub>🤖 Adversarial review by 2 independent Opus agents. Security findings verified by 10 adversarial probes plus reproduction of the author's revert experiment; the `Display` gap, CI status, and jsonwebtoken internals re-verified independently before posting.</sub>

marcjazz (approved):

Issue #246, opened 2026-08-24
authkestra-actix: Auth<I> extractor drops all request extensions except ClientCertificateDer


stephane-segning:
## Summary

\`authkestra-actix\`'s generic \`Auth<I>::from_request\` impl hand-builds an \`http::request::Parts\` via \`http::Request::builder()\` (\`crates/authkestra-actix/src/lib.rs\`, around line 312-316). That builder only carries over method/URI/headers — **every actix request extension is dropped** when it's used to build the parts handed to \`AuthenticationStrategy::authenticate\`.

## What was fixed, and what wasn't

PR #231 (RFC 8705 cert-bound tokens) added a targeted fix that carries exactly one extension type across:

\`\`\`rust
// crates/authkestra-actix/src/lib.rs:337
if let Some(cert) = req_clone.extensions().get::<ClientCertificateDer>() {
    parts.extensions.insert(cert.clone());
}
\`\`\`

This closes the gap for \`ClientCertificateDer\` specifically (needed for \`require_cert_binding\`), but the underlying defect is general. \`AuthenticationStrategy<I>\` is a **public trait** meant for downstream extension (\`crates/authkestra-engine/src/auth/strategy.rs\`) — any custom strategy a host application writes that reads a *different* extension type (tenant id, session context, a \`DeviceIdentity\` inserted by other middleware, etc.) will still silently see \`None\` under actix. The equivalent axum code path never had this problem — axum's \`FromRequestParts\` receives the framework's own native \`&mut http::request::Parts\` directly, so nothing is ever dropped there.

There is currently no live internal regression, since \`ClientCertificateDer\` is the only extension type this repo's own shipped strategies consume this way — but any downstream user of a custom \`AuthenticationStrategy\` hits this today.

## Also noted

\`authkestra-actix\` has **no test files at all** (\`git grep\` for \`#[test]\`/\`#[tokio::test]\` in that crate returns nothing). The targeted \`ClientCertificateDer\` fix shipped with no regression test, because there's no harness to put one in.

## Suggested fix

Either:
1. Carry over the full extension map generically (if \`http::Extensions\` supports iteration/cloning in a way that permits this — needs checking), or
2. Provide a documented, explicit mechanism for a host to register which extension types must survive into \`Parts\` (a small registry or a generic parameter), rather than hardcoding one type by name.

Either way, this needs its own regression test, which likely means standing up a minimal test harness for \`authkestra-actix\` first.

## References

[...]

PR #252, opened 2026-08-27
fix(actix): let request extensions reach AuthenticationStrategy

stephane-segning:
Refs #246. Deliberately **not** auto-closing: this ships option 2 from the issue (explicit registration), not option 1 (automatic whole-map carry-over), because option 1 is not expressible on the pinned versions. The maintainer should decide whether that satisfies the issue.

## The defect

`Auth<I>::from_request` synthesises an `http::request::Parts` from the request's method, URI and headers, so the extension map handed to `AuthenticationStrategy::authenticate` started empty. #231 patched this for `ClientCertificateDer` only. `AuthenticationStrategy<I>` is a public trait meant for downstream extension, so any custom strategy reading a different extension type (tenant id, session context, a `DeviceIdentity` from other middleware) still silently saw `None` under actix — though not under axum, whose `FromRequestParts` receives native `Parts`.

## Why the whole map cannot just be copied

I checked this on the pinned versions (`actix-http` 3.13.3, `actix-web` 4.14.0, `http` 1.5.0) rather than assuming. Three independent blockers, any one of which is fatal:

1. **No enumeration.** `actix_http::Extensions` wraps a *private* `HashMap<TypeId, Box<dyn Any>>` (`actix-http-3.13.3/src/extensions.rs:35`) and exposes only typed accessors — `get::<T>`, `get_mut::<T>`, `remove::<T>`, `insert::<T>`, `clear`, `extend(Extensions)`. There is no `iter()`, no `drain()`, no `IntoIterator`. Nothing outside actix-http can discover which types a request is carrying.
2. **Bounds are unrecoverable after erasure.** Even given enumeration, actix stores `Box<dyn Any>`, while `http::Extensions` stores `Box<dyn AnyClone + Send + Sync>` and its only insertion API is `insert<T: Clone + Send + Sync + 'static>(val: T)` (`http-1.5.0/src/extensions.rs:6,62`). A `Box<dyn Any>` cannot be re-boxed into that.
3. **Two different `http` crates.** actix-http 3.x depends on `http` 0.2.7, so its `Extensions` and `http` 1.x's `Extensions` are unrelated types from separately-compiled crates; `extend` has nothing to bridge.

Storing a *handle* to actix's map inside the `Parts` instead is blocked too: `HttpRequest` and `Extensions` are `!Send + !Sync`, and `http::Extensions::insert` requires `Send + Sync`.

## What this ships instead

`CarriedExtensions`: an `app_data`-registered list of extension *types* to copy across.

```rust
App::new()
    .wrap_fn(|req, srv| {
        req.extensions_mut().insert(TenantId("acme".to_string()));
        srv.call(req)
    })
    .app_data(web::Data::new(
        CarriedExtensions::new().carry::<TenantId>(),
    ))
```

Registration is generic over `T: Clone + Send + Sync + 'static` and monomorphises one `fn` pointer per type, so **the adapter no longer names any concrete extension type** — the host does. That is the sense in which this is general; it is not automatic. The `Clone + Send + Sync` bounds are not a design choice, they are exactly what `http::Extensions::insert` demands.

Both `web::Data::new(registry)` and the bare `.app_data(registry)` form are honoured, since matching only one would turn the other into the same class of silent no-op this issue is about.

`ClientCertificateDer` is still carried unconditionally and needs no registration, so the RFC 8705 resource-side check (#224) cannot be broken by a host that forgets to register it.

The full rationale lives in the `extensions` module docs so the next reader does not have to re-derive it from the registry sources.

## Tests

Five new integration tests in `crates/authkestra-actix/tests/auth_extensions_integration.rs`, driving a real `actix_web::App` via `test::call_service` like the crate's existing integration tests, plus three unit tests in the new module and one doctest.

The load-bearing test uses `TenantId`, a **test-only marker type, not `ClientCertificateDer`** — a test written against `ClientCertificateDer` would have passed on the buggy code and proved nothing.

I verified that claim rather than asserting it. Neutering only the carry-over call (keeping the API so the test still compiles, reproducing `main`'s runtime behaviour exactly):

```
test custom_extension_type_reaches_a_custom_strategy ... FAILED
test registry_also_works_without_web_data_wrapper ... FAILED

---- custom_extension_type_reaches_a_custom_strategy stdout ----
panicked at crates/authkestra-actix/tests/auth_extensions_integration.rs:109:5:
assertion `left == right` failed: TenantId inserted by middleware did not reach the strategy
  left: 401
 right: 200

test result: FAILED. 3 passed; 2 failed
```

Note which test is in the *passing* three: `client_certificate_der_is_still_carried_without_registration`. That is the direct evidence that a `ClientCertificateDer`-only test does not cover this fix.

With the fix restored, all five pass, as do the crate's 8 pre-existing integration tests.

## Verification run

```
cargo fmt --all -- --check                                                  # clean
cargo +1.98.0 clippy -p authkestra-actix --all-targets --all-features -- -D warnings   # clean
cargo test -p authkestra-actix --all-features -- --test-threads=2           # 3 + 5 + 5 + 3 + 1 doctest, all pass
```

Scoped to the touched crate rather than `--workspace`: the machine was under heavy concurrent load from sibling agents and a separate project. An earlier full-workspace run on this branch surfaced two failures, both traced to local port contention and neither reachable from this change — `authkestra`'s `examples_e2e` (port 3000 was held by an orphaned `axum_oauth2_git` from another worktree, pid 2142634) and `authkestra-engine`'s `test_mysql_indexed_store` (rootless-docker `RootlessKit PortManager.AddPort ... address already in use`). CI runs the full workspace suite on a clean machine.

## Not done

`docs/book/ch06-adapters-and-integrations.md` was left alone; the new mechanism is documented in the crate README and module docs only.

### DIFF
--- a/crates/authkestra-actix/Cargo.toml
+++ b/crates/authkestra-actix/Cargo.toml
@@ -68,3 +68,7 @@ required-features = ["macros", "session"]
 [[test]]
 name = "engine_token_integration"
 required-features = ["macros", "token"]
+
+[[test]]
+name = "auth_extensions_integration"
+required-features = ["resource"]
--- a/crates/authkestra-actix/README.md
+++ b/crates/authkestra-actix/README.md
@@ -79,6 +79,42 @@ async fn external_api(Jwt(claims): Jwt<MyClaims>) -> HttpResponse {
 }
 ```
 
+#### `Auth<I>` and request extensions
+
+`Auth<I>` runs the `Guard<I>` from `app_data` over the request. actix has no
+`http::request::Parts` of its own (actix-http 3.x is still built on `http` 0.2),
+so the adapter synthesises one from the request's method, URI and headers — and
+that synthetic value starts with an **empty extension map**.
+
+If your `AuthenticationStrategy` reads a request extension put there by another
+middleware, register its type with `CarriedExtensions`:
+
+```rust
+use actix_web::{dev::Service as _, web, App, HttpMessage};
+use authkestra_actix::CarriedExtensions;
+
+#[derive(Clone)]
+struct TenantId(String);
+
+let app = App::new()
+    .wrap_fn(|req, srv| {
+        req.extensions_mut().insert(TenantId("acme".to_string()));
+        srv.call(req)
+    })
+    .app_data(web::Data::new(
+        CarriedExtensions::new().carry::<TenantId>(),
+    ));
+```
+
+The strategy then reads `parts.extensions.get::<TenantId>()`.
+
+Registration is required because `actix_http::Extensions` exposes no way to
+enumerate its contents and stores `Box<dyn Any>`, while `http::Extensions`
+accepts only `T: Clone + Send + Sync`; copying the whole map generically is not
+expressible. See the `extensions` module docs for the full rationale.
+`ClientCertificateDer` is carried unconditionally and needs no registration, so
+RFC 8705 certificate-bound token checks keep working out of the box.
+
 ### OAuth2 Helpers
 
 The crate provides helpers to manage the OAuth2 flow lifecycle.
--- a/crates/authkestra-actix/src/extensions.rs
+++ b/crates/authkestra-actix/src/extensions.rs
@@ -1,0 +1,220 @@
+//! Carrying actix request extensions into the `http::request::Parts` handed to
+//! [`AuthenticationStrategy::authenticate`].
+//!
+//! # Why this module exists at all
+//!
+//! `authkestra-engine`'s [`AuthenticationStrategy`] is defined over
+//! [`http::request::Parts`] (`http` 1.x). axum hands its extractors that exact
+//! type, so `authkestra-axum` passes the framework's own `Parts` straight
+//! through and nothing is ever lost. actix-web has no such type: its request is
+//! an `actix_http::RequestHead` plus a separate `RefCell<Extensions>` map, and
+//! actix-http 3.x is still built on `http` **0.2**. So the actix adapter has to
+//! *synthesise* an `http` 1.x `Parts` from method + URI + headers, and that
+//! synthetic value starts with an empty extension map (issue #246).
+//!
+//! # Why the whole map cannot simply be copied
+//!
+//! A fully generic "move/clone every extension across" is not expressible on the
+//! pinned versions (`actix-http` 3.13, `http` 1.5). Three independent blockers,
+//! any one of which is fatal:
+//!
+//! 1. **No enumeration.** `actix_http::Extensions` wraps a private
+//!    `HashMap<TypeId, Box<dyn Any>>` and exposes only *typed* accessors
+//!    (`get::<T>`, `remove::<T>`, `extend(Extensions)`). There is no `iter()`,
+//!    no `drain()`, no `IntoIterator`. Nothing outside actix-http can discover
+//!    which types a request is carrying.
+//! 2. **No `Send`/`Sync`/`Clone` on the values.** Even given enumeration, actix
+//!    stores `Box<dyn Any>`, while `http::Extensions` stores
+//!    `Box<dyn AnyClone + Send + Sync>` and its only insertion API is
+//!    `insert<T: Clone + Send + Sync + 'static>(val: T)`. A type-erased
+//!    `Box<dyn Any>` cannot be re-boxed into that; the bounds are unrecoverable
+//!    once the concrete type is gone.
+//! 3. **Two different `http` crates.** `http` 0.2's `Extensions` and `http`
+//!    1.x's `Extensions` are unrelated types from separately-compiled crates,
+//!    so even `extend` has nothing to bridge.
+//!
+//! Storing a handle to actix's map inside the `Parts` instead is equally
+//! blocked: `HttpRequest`/`Extensions` are `!Send + !Sync`, and
+//! `http::Extensions::insert` requires `Send + Sync`.
+//!
+//! # What this module does instead
+//!
+//! Option 2 from issue #246: an explicit, host-registered list of extension
+//! *types* to carry. Registration is generic over `T` and monomorphises a
+//! transfer function per type, so the adapter never names a concrete extension
+//! type — the host decides. See [`CarriedExtensions`].
+//!
+//! [`AuthenticationStrategy`]: authkestra_engine::auth::AuthenticationStrategy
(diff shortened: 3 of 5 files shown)


stephane-segning:
## Adversarial review

Lens: **is the "impossible" claim actually true, is the alternative sound, is the test rigorous**. The reviewer was told explicitly that "it can't be done" is also what someone says when they didn't find the way — and to go looking for the way.

**Verdict: sound. The impossibility argument survived, the design is genuinely generic, and the neutered-state proof reproduced verbatim.** No blocking findings. Six items raised, all fixed in `32b701e`.

---

### ✅ The impossibility claim holds — and is over-determined

Verified against registry sources, not inferred from absence:

1. **No enumeration.** `actix-http-3.13.3/src/extensions.rs:33-36` — private `HashMap<TypeId, Box<dyn Any>>`. The complete public API (lines 38–186) is `new`, `insert`, `contains`, `get`, `get_mut`, `get_or_insert`, `get_or_insert_with`, `remove`, `clear`, `extend`. No `iter`, `drain`, `IntoIterator`, `keys`, or `len`. Even the `Debug` escape hatch is closed: `:188-192` is `debug_struct("Extensions").finish()`, printing nothing.
2. **Bounds unrecoverable — and *stronger* than the PR states.** `http-1.5.0/src/extensions.rs:6,62` requires `Box<dyn AnyClone + Send + Sync>`, and `trait AnyClone: Any` at `:288` is **private**. No external crate can construct that box type at all, even given enumeration and bounds.
3. **Two `http` crates.** `actix-http-3.13.3/Cargo.toml:290-291` requires `http = "0.2.7"`. (The PR quotes the manifest requirement; `0.2.12` is what resolves — no imprecision.)

Blockers 1 and 2 are each independently fatal, so the argument is over-determined — a strength, not redundancy.

**Routes the reviewer tried and closed:** `extensions_mut()` + `mem::take` (owned map still can't be enumerated, and is `!Send + !Sync + !Clone`); `Extensions::extend` (actix→actix only); a thread-local scope guard (would force the shared trait to reach into actix-specific state, violating `AGENTS.md:6`, and breaks on axum's multi-threaded runtime). `actix-web` adds nothing — `dev.rs:12` is a plain re-export.

**One alternative it did surface, offered as a supplement not a refutation:** `http::Extensions` is itself `Clone + Default` with `Send + Sync` values, so a newtype wrapping one satisfies both actix's `'static` bound and `http`'s insert bounds — letting a host carry arbitrarily many types with zero per-type registration. But it's *worse* for this issue's actual use case (an extension inserted by a **third-party** middleware into actix's own map), since it requires that middleware to write into authkestra's parallel map instead. Worth knowing about; doesn't displace the registry.

### ✅ Design and proof

- `carry::<T>()` is genuinely generic — `extensions.rs:129` pushes `carry_one::<T>` over a `fn` pointer type. No hardcoded list. The adapter names exactly one concrete type, at `lib.rs:350`.
- `lib.rs:301`/`:316` is the **only** synthetic-`Parts` construction site in either adapter, and `:383` the only `authenticate` call — so the fix is complete for this defect class.
- Unconditional `ClientCertificateDer` is justified: without it `require_cert_binding` fails *closed*, so a forgotten registration would be an availability bug, not a security hole.
- No locking, no per-request allocation, no `RefCell` borrow-panic risk. The `Ref` never crosses an await, and a re-entrant `Clone` impl can't exist because it would need to capture `!Send + !Sync` types that can't satisfy `carry::<T>`'s bounds.
- **Neutering `registry.apply(...)` reproduced the claimed output exactly** — `custom_extension_type_reaches_a_custom_strategy` and `registry_also_works_without_web_data_wrapper` fail 401-vs-200 while `client_certificate_der_is_still_carried_without_registration` **passes**. Confirms a `ClientCertificateDer`-only test would have proved nothing.
- 401 is a specific signal, not a catch-all: `Ok(None)` → 401, `Err` → 500, missing `Guard` → 500. Positive tests also assert the body.
- **The new-test-file decision was verified empirically.** Adding `resource` to `engine_token_integration`'s `required-features` and running `--features macros,token --no-run` made that target **silently disappear** — exit 0, no warning. The reasoning was exactly right, and the failure would have been the silent kind.

---

### 🔧 Fixed in `32b701e`

**1. The new mechanism's own failure mode was still silent.** Both arms logged at `debug!`, which is off in most production setups — so a host registering the wrong type, or ordering middleware after the extractor, still got a bare 401 with no signal. `extensions.rs` even named the problem ("*which is otherwise completely silent*") and then logged it invisibly. Now `warn!` when types are registered but nothing was carried. The no-registry arm is demoted to `trace!`, since it fired on every request of every app that never opts in.

**2. `web::Data` precedence was undocumented.** `HttpRequest::app_data` scans the entire App → Scope → Resource chain for one type *before* the `or_else` fallback runs, so an app-level `web::Data` registration beats a resource-level bare one. Accepting both forms is right; the ordering between them is now documented.

**3. Incorrect justification comment.** The `Carrier` doc claimed `app_data` "requires `Send + Sync`". It requires only `'static` — those bounds arrive transitively from `HttpServer::new`'s factory. The `fn`-pointer choice is still right; the stated reason wasn't.

**4. Feature gate undocumented.** `CarriedExtensions` is `resource`-gated; a reader on `--features token` got an unresolved import with no hint. Now stated.

**5. Coverage gap — no test registered two types both present.** Every positive test registered types of which only one was actually inserted, so a registry that stopped after its first carrier would have passed all of them. Added `two_registered_types_are_both_carried`, and **verified it load-bearing**: restricting `apply()` to `.take(1)` fails that test alone while the other five keep passing.

**6.** Also fixed a `clippy::len_zero` I introduced while making fix 1 (`len() > 0` → `!is_empty()`), caught by running the CI-toolchain clippy rather than assuming.

---

### 📋 Noted, not actioned

- **Commit type.** Single commit is `fix(actix):`, but this adds a new public type. Additive API in a patch is semver-legal, and release-plz drives the changelog from the commit type — `feat(actix):` would describe it better. Left alone since retyping the commit is your call.
- **README example is never compiled** — `lib.rs` has no `#![doc = include_str!("../README.md")]`. The equivalent `extensions.rs` doctest *is* compiled and passes.
- **The two negative tests are contract pins, not regression detectors** — no mutation of this change can make them fail. Worth having, but don't count them as regression coverage.

<sub>🤖 Adversarial review by an Opus agent. Impossibility claim checked against registry sources with four alternative routes attempted; the neutered-state proof and the feature-gating decision both reproduced empirically. The two-type coverage fix was independently mutation-verified before posting.</sub>

marcjazz (approved):

marcjazz:
Should close #246

Issue #242, opened 2026-08-24
EdDSA verification uses non-strict ed25519_dalek::Verifier::verify, not verify_strict — small-order keys defeat proof-of-possession in authkestra-devsig


stephane-segning:
## Summary

Every `authkestra-*` crate verifies EdDSA through `jsonwebtoken`'s `rust_crypto` backend, which calls `ed25519_dalek::Verifier::verify` — the **non-strict** verifier. It never calls `verify_strict`. As a result authkestra accepts signatures under small-order (including identity) public keys that strict verification rejects.

## Why this matters — proof-of-possession collapses

A small-order public key has no meaningful secret: an all-zero key with an all-zero signature verifies against **any** message. Under non-strict verification, "this signature proves the holder possesses the private key" stops being true — anyone can produce a valid signature for that identity.

That is mostly theoretical where the verifying key comes from a trusted JWKS the attacker cannot influence (`authkestra-resource`'s JWT path). It is **not** theoretical in `authkestra-devsig`, where the public key travels in the request itself: `authkestra-devsig`'s `X-Signature` check (`crates/authkestra-devsig/src/signature.rs`, `crypto::verify(...)`) goes through the identical non-strict path. If a small-order key can be enrolled and attested, every subsequent `X-Signature` under that identity is forgeable by any party — which defeats the per-request proof-of-possession the crate exists to provide.

## Evidence

- `jsonwebtoken` 11.0.0, `src/crypto/rust_crypto/eddsa.rs` — the EdDSA verifier calls `Verifier::verify`, not `verify_strict`.
- `ed25519-dalek` 2.2.0, `src/verifying.rs` — `verify_strict` rejects small-order `R` and small-order public keys via `is_small_order()` before the verification equation; the non-strict `raw_verify` path performs no such check.
- Both versions confirmed as the real resolved graph via `cargo tree -i ed25519-dalek` against two deployed consumers whose lockfiles pin the authkestra family at 0.5.4.
- Empirically demonstrated in a test: the all-zero low-order key/signature pair is passed to `jsonwebtoken::crypto::verify` with `Algorithm::EdDSA` and returns `Ok(true)`.

## Suggested fix

[...]

PR #253, opened 2026-08-27
fix(devsig): verify EdDSA strictly, rejecting low-order Ed25519 keys

stephane-segning:
Refs #242.

## The defect, reproduced

`jsonwebtoken` 11.0.0's `rust_crypto` backend verifies EdDSA with `ed25519_dalek::Verifier::verify`, the **non-strict** verifier (`jsonwebtoken-11.0.0/src/crypto/rust_crypto/eddsa.rs`: `self.0.verify(msg, &Signature::from_slice(signature)?)`). Non-strict verification never asks whether the public key — or the signature's `R` — is a low-order point, so the triple `(A = identity, R = identity, S = 0)` satisfies `[S]B - [k]A == R` for **every** message. No private key exists for such a key.

In `authkestra-devsig` the verifying key is not trusted input: it travels in the request, inside `X-Signature`'s protected header as an embedded `jwk`. A low-order key thumbprints exactly like a real one, so an attacker who gets one enrolled passes the `cnf.jkt` binding check honestly and then needs no secret at all to sign.

I wrote the test first and ran it against unmodified `main` (`2731e04`), source untouched:

```
test low_order_ed25519_device_key_cannot_forge_a_signature ... FAILED

thread 'low_order_ed25519_device_key_cannot_forge_a_signature' panicked at
crates/authkestra-devsig/tests/eddsa_strict.rs:58:22:
a low-order Ed25519 public key has no private key behind it -- accepting a signature under it
means anyone can sign as this identity, which defeats the per-request proof-of-possession this
crate exists to provide (authkestra#242): DeviceIdentity { subject: "user-7f3a9c12",
device: "dev-8821bc04", key_thumbprint: "eV9frzBXPTP92MWWMpoFOh0WI_kJLvGlhcNs15APU_s",
attributes: Object {"kyc_level": Number(2), "roles": Array [String("customer")],
"status": String("active")} }
```

`verify()` returned `Ok(DeviceIdentity { .. })`. Every other check in the algorithm passed honestly — the attestation really is the issuer's, the embedded `jwk` really does thumbprint-match `cnf.jkt`, the claims really do bind the method/path/audience, the signature is fresh and unreplayed. Only the EdDSA check stood between that request and a full impersonation, and it waved it through. This is a complete authentication bypass.

The same run showed the issuer side is exposed the same way if a low-order key ever reaches the cached JWKS:

```
test low_order_ed25519_issuer_key_cannot_forge_an_attestation ... FAILED
... an attestation 'signed' by a low-order Ed25519 issuer key must be rejected -- accepting it
lets anyone who can influence the JWKS mint attestations for any subject:
DeviceIdentity { subject: "user-7f3a9c12", device: "dev-8821bc04", ... }
```

### One correction to the issue's wording

#242 says "an all-zero key with an all-zero signature verifies against **any** message". The all-zero encoding is low-order, but it is the order-4 point `y = 0`, and the equation only holds for it when the challenge scalar `k ≡ 3 (mod 4)` — roughly one message in four, so it is not a deterministic test vector. The **identity** encoding (`01 00 … 00`) is the universal one: both sides of the equation collapse to the identity whatever `k` is. The tests use that. The reasoning and the conclusion in the issue are otherwise exactly right.

## The fix

New `crates/authkestra-devsig/src/eddsa.rs`, used by both `signature.rs` and `attestation.rs`:

1. **`VerifyingKey::verify_strict`** replaces `jsonwebtoken::crypto::verify` on the EdDSA branch. This also rejects a low-order `R`.
2. **Low-order keys are rejected outright** via `is_weak()`, *before* verification is attempted — so the key is refused on its own merits, not depending on which signature bytes came with it. Surfaced as `bad_jwk`, not `bad_signature`: the key is the defect.

Either layer alone blocks acceptance; the second exists so the rejection is correctly classified and independent of the accompanying signature. Both were confirmed load-bearing by deleting them and re-running (see below).

Every non-EdDSA algorithm still delegates to `jsonwebtoken` unchanged — no RSA or ECDSA verification is reimplemented.

After:

```
running 32 tests
test canary_upstream_jsonwebtoken_still_verifies_eddsa_non_strictly ... ok
test case_01_valid_pair_is_accepted ... ok
...
test genuine_ed25519_device_key_is_still_accepted ... ok
test genuine_ed25519_device_key_with_a_tampered_body_is_rejected ... ok
test genuine_ed25519_device_key_with_a_wrong_signature_is_bad_signature ... ok
test low_order_ed25519_device_key_cannot_forge_a_signature ... ok
test low_order_ed25519_issuer_key_cannot_forge_an_attestation ... ok
test low_order_ed25519_device_key_is_rejected_regardless_of_signature_bytes ... ok

test result: ok. 32 passed; 0 failed; 0 ignored
```

No existing test was changed to accommodate this. The pre-existing conformance cases still pass as written; the only edits to `TestSetup` were adding `EdDSA` to `allowed_algs` and threading a signer through the builders so a "key" with no private half can be expressed.

### The decisive check

Neutering just the `is_weak()` branch (`if false && key.is_weak()`) and re-running:

```
test low_order_ed25519_device_key_cannot_forge_a_signature ... FAILED
  left: "bad_signature"
 right: "bad_jwk"
test low_order_ed25519_device_key_is_rejected_regardless_of_signature_bytes ... FAILED
```

Worth reading precisely: with the key check removed the forgery is still *rejected* (`verify_strict` catches the low-order `R`), but as `bad_signature`. Both layers are doing real work.

## Tests added

| test | what it pins |
| --- | --- |
| `low_order_ed25519_device_key_cannot_forge_a_signature` | the reported bypass, end to end through `verify()` |
| `low_order_ed25519_device_key_is_rejected_regardless_of_signature_bytes` | the key is rejected on its own merits, not by matching one canned forgery vector |
| `genuine_ed25519_device_key_is_still_accepted` | **positive control** — a real Ed25519 device key still verifies, body hash included |
| `genuine_ed25519_device_key_with_a_tampered_body_is_rejected` | the strict path still actually verifies |
| `genuine_ed25519_device_key_with_a_wrong_signature_is_bad_signature` | garbage bytes stay `bad_signature`, not reclassified |
| `low_order_ed25519_issuer_key_cannot_forge_an_attestation` | JWKS-ingest hardening |
| `canary_upstream_jsonwebtoken_still_verifies_eddsa_non_strictly` | **deliberate canary**: asserts the upstream behaviour being worked around. If `jsonwebtoken` ever adopts `verify_strict`, this fails and tells us the local path can be revisited, so the workaround cannot quietly outlive its reason. |

## Behaviour change

Signatures and attestations under low-order Ed25519 keys are **now rejected where they were previously accepted**. No honestly generated key is affected — a low-order point is not something a real Ed25519 keygen produces — and no public API changed.

I did **not** mark this `!`. No consumer API breaks, and the only traffic that stops being accepted is forged. But that is a release-semantics call for the maintainer, and `release-plz` derives the bump from this prefix: **say the word and I'll amend to `fix(devsig)!`** if you'd rather the version bump advertise the behaviour change.

## Dependency

`ed25519-dalek = "2"` becomes a direct dependency of `authkestra-devsig` (production), plus `features = ["pkcs8"]` in dev-dependencies for minting a real Ed25519 test key. It adds **nothing** to the graph — `jsonwebtoken`'s `rust_crypto` backend, already enabled, resolves the same 2.2.0. The whole `Cargo.lock` diff is one line:

```
 crates/authkestra-devsig v0.5.5
 dependencies = [
   "async-trait",
   "base64 0.22.1",
+  "ed25519-dalek",
```

`cargo deny check licenses bans` → `bans ok, licenses ok`.

## Not covered here — needs a follow-up

`authkestra-resource`'s JWT path goes through the same non-strict backend. I agree with the issue that it is **mostly theoretical**: the verifying key there is resolved from a trusted OP JWKS the attacker cannot influence, and unlike devsig there is no enrolment channel through which an attacker submits a key of their choosing. The realistic prerequisite is a compromised or hostile OP, at which point strict verification is not the thing protecting you. It should still be hardened for the same defence-in-depth reason the attestation path was, but `crates/authkestra-resource/src/jwt.rs` has two PRs in flight right now, so I have deliberately **not** touched it.

For that reason there is **no auto-closing keyword** on #242 — leave it open until `authkestra-resource` is hardened too, or split that half into its own issue.

## Verification

```
cargo fmt --all -- --check                                                             # clean
cargo +1.98.0 clippy -p authkestra-devsig --all-targets --all-features -- -D warnings  # clean
cargo test -p authkestra-devsig --all-features                                         # 32 passed, 0 failed
cargo deny check licenses bans                                                         # ok
```

Scoped to `authkestra-devsig`, the only crate touched. No public API of the crate changed, so the `devsig` features in `authkestra-axum`/`authkestra-actix` are unaffected; workspace-wide CI covers them.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -528,6 +528,7 @@ version = "0.5.5"
 dependencies = [
  "async-trait",
  "base64 0.22.1",
+ "ed25519-dalek",
  "jsonwebtoken",
  "p256",
  "rand 0.8.7",
--- a/crates/authkestra-devsig/Cargo.toml
+++ b/crates/authkestra-devsig/Cargo.toml
@@ -12,6 +12,11 @@ readme = "README.md"
 
 [dependencies]
 async-trait = "0.1"
+# Direct, not transitive-by-accident: `crate::eddsa` calls `VerifyingKey::verify_strict` itself
+# because `jsonwebtoken`'s EdDSA backend uses the non-strict verifier (authkestra#242). This adds
+# no new code to the dependency graph — `jsonwebtoken`'s `rust_crypto` backend, already enabled
+# above, resolves the same `ed25519-dalek` 2.x.
+ed25519-dalek = "2"
 jsonwebtoken = { version = "11", features = ["rust_crypto"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
@@ -27,6 +32,10 @@ tokio = { version = "1", features = ["full"] }
 rsa = "0.9"
 p256 = { version = "0.13", features = ["pem", "ecdsa"] }
 rand = "0.8"
+# `pkcs8` only, and only for tests: the conformance suite needs to mint a *real* Ed25519 device
+# key in the PKCS#8 DER shape `jsonwebtoken::EncodingKey::from_ed_der` expects, so that the
+# positive EdDSA case exercises the same signing path a real device would.
+ed25519-dalek = { version = "2", features = ["pkcs8"] }
 
 [features]
 default = []
--- a/crates/authkestra-devsig/src/attestation.rs
+++ b/crates/authkestra-devsig/src/attestation.rs
@@ -1,11 +1,12 @@
 //! Parsing and trust verification of `X-Attestation`.
 
 use jsonwebtoken::jwk::Jwk;
-use jsonwebtoken::{crypto, Algorithm, DecodingKey};
+use jsonwebtoken::{Algorithm, DecodingKey};
 use serde::Deserialize;
 use serde_json::Value;
 
 use crate::config::VerifierConfig;
+use crate::eddsa::{verify_jws_signature, VerifyFailure};
 use crate::error::VerifyError;
 use crate::jwks::IssuerJwks;
 use crate::jws_util::{decode_json_segment, parse_and_check_alg, split_compact, RawJws};
@@ -96,14 +97,32 @@ pub async fn verify_trust(
     let decoding_key = DecodingKey::from_jwk(&jwk)
         .map_err(|e| VerifyError::BadAttestation(format!("issuer key unusable: {e}")))?;
 
+    // Strict EdDSA here too. The issuer key comes from the cached JWKS rather than from the
+    // request, so this is defence in depth rather than a live exposure — but a low-order key that
+    // reached the JWKS would let anyone mint attestations for any subject, and the check is free
+    // now that `crate::eddsa` exists. See that module's docs (authkestra#242).
     let signing_input = format!("{}.{}", parsed.raw.header_b64, parsed.raw.payload_b64);
-    let ok = crypto::verify(
+    let ok = verify_jws_signature(
+        parsed.alg,
+        &jwk,
+        &decoding_key,
         parsed.raw.signature_b64,
         signing_input.as_bytes(),
-        &decoding_key,
-        parsed.alg,
     )
-    .map_err(|e| VerifyError::BadAttestation(format!("signature verification error: {e}")))?;
+    .map_err(|failure| match failure {
+        VerifyFailure::Key(reason) => {
+            tracing::warn!(
+                target: "authkestra_devsig",
+                iss = %raw_claims.iss,
+                kid,
+                "attestation rejected: the issuer key resolved from the JWKS is unusable or unsafe"
+            );
+            VerifyError::BadAttestation(format!("issuer key unusable: {reason}"))
+        }
+        VerifyFailure::Backend(reason) => {
+            VerifyError::BadAttestation(format!("signature verification error: {reason}"))
+        }
+    })?;
     if !ok {
         return Err(VerifyError::BadAttestation(
             "signature does not verify".to_string(),
--- a/crates/authkestra-devsig/src/eddsa.rs
+++ b/crates/authkestra-devsig/src/eddsa.rs
@@ -165,3 +165,64 @@ fn ed25519_verifying_key(jwk: &Jwk) -> Result<VerifyingKey, VerifyFailure> {
 
     Ok(key)
 }
+
+#[cfg(test)]
+mod tests {
+    use ed25519_dalek::{Signature, Signer, SigningKey, Verifier, VerifyingKey};
+
+    /// The identity point: the canonical universal low-order vector.
+    const IDENTITY: [u8; 32] = {
+        let mut b = [0u8; 32];
+        b[0] = 1;
+        b
+    };
+
+    /// Pins `verify_strict` over the non-strict `Verifier::verify`.
+    ///
+    /// Raised in review of #242: swapping `verify_strict` for `Verifier::verify`
+    /// in [`super::verify_jws_signature`] leaves the whole conformance suite
+    /// green, so nothing pinned the strict call itself. That is expected rather
+    /// than alarming — [`super::ed25519_verifying_key`]'s `is_weak()` gate
+    /// rejects a small-order `A` *before* verification runs, and producing a
+    /// small-order `R` that satisfies the equation under a non-small-order `A`
+    /// would require solving a discrete log. So there is no reachable
+    /// end-to-end input that distinguishes the two verifiers.
+    ///
+    /// The distinguishing input therefore has to be a small-order *key*, which
+    /// means calling dalek directly here, below that gate. This is the pair the
+    /// two verifiers genuinely disagree about, and it is what makes the strict
+    /// call load-bearing rather than decorative: without it, a future refactor
+    /// could drop `verify_strict` and CI would stay green.
+    #[test]
+    fn strict_verification_rejects_a_low_order_key_that_non_strict_accepts() {
+        let weak = VerifyingKey::from_bytes(&IDENTITY).expect("identity is a valid Edwards point");
+        assert!(weak.is_weak(), "the identity point must be low-order");
+
+        // (R = identity, S = 0): both sides of [S]B - [k]A == R collapse to the
+        // identity for every challenge scalar, so this verifies under ANY message.
+        let mut forged = [0u8; 64];
+        forged[..32].copy_from_slice(&IDENTITY);
+        let forged = Signature::from_bytes(&forged);
+
+        assert!(
+            weak.verify(b"authkestra", &forged).is_ok(),
+            "precondition: the NON-strict verifier accepts this forgery — if this ever fails, \
+             upstream changed and this test no longer proves what it claims"
+        );
+        assert!(
+            weak.verify_strict(b"authkestra", &forged).is_err(),
+            "verify_strict must reject a low-order key that non-strict accepts"
+        );
+    }
+
+    /// Positive control: a fix that rejects everything is not a fix.
+    #[test]
+    fn strict_verification_still_accepts_a_genuine_signature() {
+        let signing = SigningKey::from_bytes(&[7u8; 32]);
+        let key = signing.verifying_key();
+        assert!(!key.is_weak());
+
+        let genuine = signing.sign(b"authkestra");
+        assert!(key.verify_strict(b"authkestra", &genuine).is_ok());
+    }
+}
--- a/crates/authkestra-devsig/src/eddsa.rs
+++ b/crates/authkestra-devsig/src/eddsa.rs
@@ -1,0 +1,167 @@
+//! Compact-JWS signature verification, with a **strict** Ed25519 path of this crate's own.
+//!
+//! ## Why this module exists ([authkestra#242](https://github.com/marcjazz/authkestra/issues/242))
+//!
+//! `jsonwebtoken` 11.0.0's `rust_crypto` backend verifies EdDSA with
+//! `ed25519_dalek::Verifier::verify` — the **non-strict** verifier (see
+//! `jsonwebtoken-11.0.0/src/crypto/rust_crypto/eddsa.rs`). Non-strict verification never asks
+//! whether the public key, or the signature's `R`, is a low-order point, so the triple
+//! `(A = identity, R = identity, S = 0)` satisfies the verification equation `[S]B - [k]A == R`
+//! for *every* message: both sides collapse to the identity whatever the challenge scalar `k` is.
+//! No private key exists for such a public key — not "is unknown", *does not exist*.
+//!
+//! That is fatal for this crate specifically, because here the verifying key is **not** trusted
+//! input: it travels in the request, inside the `X-Signature` protected header's embedded `jwk`.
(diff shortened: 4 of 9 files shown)


stephane-segning:
## Adversarial review — cryptographic correctness

**Verdict: the fix is cryptographically sound. The reviewer could not break it.** One MEDIUM finding about *residual scope* (now filed as #256), one LOW now fixed, and a second vulnerability this PR silently closes that deserved documenting.

### ✅ The vulnerability is real — reproduced independently

Throwaway worktree at `2731e04`, applying **only** this PR's `tests/` diff with source untouched:

```
test low_order_ed25519_device_key_cannot_forge_a_signature ... FAILED
  ... DeviceIdentity { subject: "user-7f3a9c12", device: "dev-8821bc04", ... }
test low_order_ed25519_issuer_key_cannot_forge_an_attestation ... FAILED
test low_order_ed25519_device_key_is_rejected_regardless_of_signature_bytes ... FAILED
test result: FAILED. 29 passed; 3 failed
```

`verify()` returns `Ok(DeviceIdentity)`. Full authentication bypass, exactly as described. 32/32 on the branch.

### ✅ The correction to #242's crypto is exactly right

Verified with an independent implementation of the Ed25519 group:

| encoding | valid point | order | non-strict verifies |
|---|---|---|---|
| `00…00` (all-zero, `y=0`) | yes | **4** | **52/200** messages |
| `01 00…00` (identity, `y=1`) | yes | 1 | **200/200** messages |

Grouped by challenge residue for the all-zero key: `{k≡0: 0/92, k≡1: 0/94, k≡2: 0/103, k≡3: 111/111}`. So "order-4 point, holds iff `k ≡ 3 mod 4`" is precise, the issue's "verifies against any message" for the all-zero pair is wrong, and the tests correctly use the identity encoding.

### ✅ Coverage, completeness, no false negatives

- **`is_weak()` covers the whole small-order subgroup** — `is_small_order` → `mul_by_cofactor().is_identity()`, i.e. `[8]P == id`, all 8 torsion points, no gaps. All 8 canonical torsion encodings × 3 signature shapes → **24/24 rejected as `bad_jwk`**. The three non-canonical libsodium-blacklist encodings likewise.
- **Complete within the crate** — exactly two verification sites existed (`signature.rs:167`, `attestation.rs:100`); both converted. No other `crypto::verify` / `decode` / `Algorithm::EdDSA` path.
- **Branch-local** — `eddsa.rs:72-74` delegates every non-EdDSA algorithm untouched; all 19 pre-existing ES256/RS256 conformance cases pass.
- **Non-canonical `S` (`S >= L`) rejected** via `check_scalar`, confirmed empirically with `S' = S + L`. `legacy_compatibility` is not enabled anywhere in the graph.
- **Dependency clean** — one `ed25519-dalek` node at 2.2.0, no duplicate majors; `pkcs8` is already in the non-dev graph via `jsonwebtoken`, so the dev-dep feature is a no-op.

---

### 🔴 MEDIUM — the "Not covered here" section understates residual exposure → filed as #256

The PR says the remaining gap is `authkestra-resource` and is "mostly theoretical — the realistic prerequisite is a compromised or hostile OP." **That is not accurate**, and I re-verified this against `main` myself before filing.

`authkestra-op`'s **enrolment proof-of-possession** verifies against a JWK that is unauthenticated request input:

- `handlers/enrolment.rs:39` — `EnrolStartRequest { pub public_jwk: Value }`
- `handlers/enrolment.rs:300-302` — `parse_public_jwk(...)` then `verify_challenge_signature(&req.challenge_signature, &jwk)`
- `attestation.rs:263-290` — `parse_public_jwk` checks private members and `kty` only; **no small-order check**
- `attestation.rs:373` — any OKP routes to `Algorithm::EdDSA`
- `attestation.rs:429` — `jsonwebtoken::decode`, the same non-strict backend

**Attack:** enrol with the identity point, complete the challenge with the canned `(R = identity, S = 0)` signature, and the OP mints a real attestation binding that `cnf.jkt`. No compromised OP required — this is precisely the precondition this PR's own threat model names ("an attacker who gets one enrolled"), and it lives in this repo.

Post-fix `authkestra-devsig` rejects those keys, so the end-to-end bypass is closed *for this verifier*. But the OP still issues attestations for keys nobody holds, exploitable against pre-#253 devsig versions, `authkestra-resource`, and third-party verifiers.

Tracked in **#256**. Lower-reachability sibling: `client_assertion.rs:388` (needs an admin-registered key; no dynamic registration exists). Leaving #242 open with no closing keyword was the right call and this strengthens it.

### 🔧 LOW — `verify_strict` wasn't pinned by any test → fixed in `a8f6188`

Swapping `verify_strict` for `Verifier::verify` left the entire 32-test suite **green**. That's expected rather than alarming — `is_weak()` rejects small-order `A` before verification, and producing a small-order `R` satisfying the equation under non-small-order `A` needs a discrete log — so no reachable end-to-end input distinguishes them. But a future refactor could silently drop `verify_strict`.

Added a unit test calling dalek directly, below the `is_weak()` gate, with the identity point and `(R = identity, S = 0)`. It asserts the *precondition* that non-strict **accepts** the forgery, then that strict rejects it — so if upstream ever adopts strict verification it fails loudly instead of quietly proving nothing. Verified load-bearing: swapping the assertion to `Verifier::verify` fails it.

Worth recording: my first attempt at this test was **itself vacuous** — it used a non-small-order key, where the forgery fails the equation under both verifiers, so it passed either way. Same defect class the finding was about.

Also fair to note: the PR's "both were confirmed load-bearing" is true for **error classification**, which is what was actually demonstrated, but not in the security sense. The body read slightly stronger than the evidence.

### 📌 INFO — this PR silently fixes a *second* vulnerability

On `main`, an `X-Signature` embedded jwk with `kty=OKP, crv=Ed25519` and an `x` shorter than 32 bytes **panics**:

```
panicked at jsonwebtoken-11.0.0/src/crypto/rust_crypto/eddsa.rs:48:72:
range end index 32 out of range for slice of length 0
```

Unauthenticated, attacker-controlled, and *outside* `signature.rs`'s `catch_unwind` (which wraps only `jwk.thumbprint()`, not `crypto::verify`). `ed25519_verifying_key` validates length first (`eddsa.rs:139-145`); `x` lengths `{0,1,16,31,33,64}` all return `bad_jwk` cleanly. **This is a second fix and belongs in the PR body.**

### 📋 Remaining, not actioned

- **L2** — no positive control for a genuine EdDSA *issuer* key (`TestSetup`'s issuer is RSA-2048), so the `attestation.rs` conversion has no regression guard for legitimate EdDSA issuers. The reviewer wrote one and it passes — no bug, just an uncovered branch.
- **INFO-2** — mixed-order keys (`A = Q + T`) are not flagged weak and verify normally. Documented dalek behaviour, not a forgery vector, but it's the edge of what this fix covers and the module docs could say so.
- **INFO-3** — `verify_jws_signature` takes both `jwk` and `decoding_key` and trusts they correspond. Both call sites derive one from the other immediately beforehand and the doc warns about it, but taking `&Jwk` alone would make the invariant unbreakable.
- **Semver** — whether to amend to `fix(devsig)!` remains a maintainer call; correctly surfaced rather than decided.

<sub>🤖 Adversarial review by an Opus agent. Bypass reproduced on `main`; the Ed25519 group arithmetic re-derived independently; all 8 torsion points and 3 non-canonical encodings exercised end-to-end. The MEDIUM finding was re-verified against `main` before filing #256.</sub>

marcjazz (approved):

stephane-segning:
## Review findings addressed

### Fixed

- **`a8f6188`** — `verify_strict` was not pinned by any test. Swapping it for the non-strict verifier left all 32 tests green, because `is_weak()` rejects small-order `A` before verification and no reachable end-to-end input distinguishes the two. Added a unit test calling dalek directly below that gate, asserting the *precondition* that non-strict **accepts** the identity-point forgery and that strict rejects it. Verified load-bearing.
  - Worth recording: my first attempt at that test was itself vacuous (non-small-order key → forgery fails the equation under both verifiers → passed either way). Same defect class the finding was about.
- **`d7be83c`** — no positive control for a genuine EdDSA **issuer** key. `TestSetup`'s issuer is RSA-2048, so the only EdDSA attestation coverage was negative; a change rejecting *every* EdDSA issuer would have passed the whole suite. Added `genuine_ed25519_issuer_key_is_still_accepted`, mirroring the forged-issuer test. 33 tests pass.

### ⚠️ This PR also fixes a second, previously undocumented vulnerability

Raised in review and worth stating plainly rather than leaving as a side effect.

On `main`, an `X-Signature` embedded jwk with `kty=OKP, crv=Ed25519` and an `x` shorter than 32 bytes **panics**:

```
panicked at jsonwebtoken-11.0.0/src/crypto/rust_crypto/eddsa.rs:48:72:
range end index 32 out of range for slice of length 0
```

Unauthenticated, attacker-controlled, and **outside** `signature.rs`'s `catch_unwind` — which wraps only `jwk.thumbprint()`, not `crypto::verify`. So a single malformed header could take down a request worker.

`ed25519_verifying_key` validates the length before touching the backend (`eddsa.rs:139-145`); `x` lengths `{0, 1, 16, 31, 33, 64}` all return `bad_jwk` cleanly on this branch.

### ⚠️ Residual scope corrected — filed as #256

The "Not covered here" section said the remaining gap was `authkestra-resource` and was "mostly theoretical — the realistic prerequisite is a compromised or hostile OP." **That understated it**, and I re-verified against `main` before filing.

`authkestra-op`'s enrolment PoP verifies against a JWK that is *unauthenticated request input* (`handlers/enrolment.rs:39`), `parse_public_jwk` has no small-order check (`attestation.rs:263-290`), and any OKP routes to the same non-strict backend (`attestation.rs:373,429`). An attacker enrols the identity point, completes the challenge with the canned `(R = identity, S = 0)` signature, and the OP mints a real attestation binding that `cnf.jkt` — no compromised OP required. That is precisely the precondition this PR's own threat model names.

Post-fix `authkestra-devsig` rejects those keys, so the end-to-end bypass is closed *for this verifier*; the OP still issues attestations for keys nobody holds. Tracked in **#256**, priority ordering: enrolment first, `client_assertion.rs` second, `authkestra-resource` last.

Leaving #242 open with no closing keyword was the right call, and this strengthens it.

### Not actioned — left for you

- **Mixed-order keys** (`A = Q + T`) are not flagged weak and verify normally. Documented dalek behaviour, not a forgery vector, but it is the edge of what this fix covers.
- **`verify_jws_signature` takes both `jwk` and `decoding_key`** and trusts they correspond. Both call sites derive one from the other immediately beforehand and the doc warns about it; taking `&Jwk` alone would make the invariant unbreakable.
- **Semver** — whether to amend to `fix(devsig)!`. The fix changes what is accepted, though only forged traffic. Still your call.

Issue #243, opened 2026-08-24
authkestra-resource: ValidationConfig is single-issuer — no per-issuer JWKS resolution for a multi-issuer resource server


stephane-segning:
## Summary

`authkestra-resource`'s `ValidationConfig` binds **one JWKS endpoint to one issuer**, so a single resource verifier cannot accept tokens from several distinct issuers. There is no trust-map or per-issuer key-resolution seam.

## Current shape

`crates/authkestra-resource/src/jwt.rs` (at `authkestra-v0.5.4`):

```rust
pub struct ValidationConfig {
    pub jwks_url: String,          // one endpoint
    pub issuer: Option<String>,    // one issuer
    ...
}
```

and validation does:

```rust
if let Some(iss) = config.issuer {
    validation.set_issuer(&[iss]);   // jsonwebtoken accepts a slice; only one is ever passed
}
```

Two distinct limitations, worth separating because they differ a lot in cost:

1. **`iss` matching is artificially narrowed.** `jsonwebtoken`'s `set_issuer` already takes a slice, so accepting several issuer names would be a small change to `ValidationConfig` (`Option<String>` → a collection) with no work in the verification path.
2. **Key resolution is the real gap.** Even with a list of accepted `iss` values, there is still exactly one `jwks_url` per config. Distinct issuers have distinct signing keys at distinct endpoints, so a genuine multi-issuer verifier needs to resolve *which* JWKS to use from the token's own `iss` — a per-issuer map, or a resolver hook. `grep` for `issuer_resolver` / `issuers` / trust-map concepts in `authkestra-resource` returns nothing at 0.5.4.

## Why it matters

A resource server that legitimately trusts more than one issuer today has to build and dispatch between several validators by hand, outside the crate — which also means the JWKS caching/refresh machinery (`refresh_interval`) is duplicated per issuer rather than managed once.

Concretely, `vaam-store/platform` maintains its own `MultiIssuerJwksVerifier` taking a `HashMap<issuer, jwks_url>` precisely so one verifier can trust several issuers at once (it backs a `*_TICKET_ISSUERS` trust map). Adopting `authkestra-resource` as-is would mean either keeping that component alongside authkestra or losing the capability.

## Suggested shape

Either:
- extend `ValidationConfig` to carry a map of `issuer -> jwks_url` (with the existing single-issuer form as the one-entry case, so no breaking change), resolving the JWKS from the validated `iss`; or
[...]

PR #254, opened 2026-08-27
feat(resource): resolve JWKS per issuer for multi-issuer resource servers

stephane-segning:
Refs #243. **Deliberately does not close it** — two design questions in it are maintainer calls and are left open below.

`ValidationConfig` bound one JWKS endpoint to one issuer, so a resource server that legitimately trusts several issuers had to build and dispatch between separate validators by hand — which also duplicated the JWKS cache/refresh machinery per issuer, the concrete complaint in the issue (`vaam-store/platform`'s `MultiIssuerJwksVerifier`).

## The seam

The issue offers two shapes ("map of `issuer -> jwks_url`" or "a resolver hook") and notes the second composes better for runtime-discovered issuers. This implements the resolver as the *mechanism* and the map as the *configured case*, so neither is a second parallel path:

| Item | Kind |
| --- | --- |
| `trait JwksResolver` — resolve a `JwksCache` from the token's `iss` (`Box<dyn>` on the I/O path, per `AGENTS.md`) | new |
| `IssuerTrustMap` — fixed `iss` -> `JwksCache` map; one cache per issuer, so TTL/rotation refresh stays per-endpoint but stays managed by this crate | new |
| `SingleJwksResolver` — one endpoint for any `iss`; the pre-existing behaviour, unchanged | new |
| `ValidationConfig::trusted_issuers: BTreeMap<String, String>` | new field |
| `ValidationConfigBuilder::trusted_issuer(iss, url)` / `trusted_issuers(pairs)` | new |
| `JwtStrategy::with_resolver(config, Box<dyn JwksResolver>)` — for issuers only known at runtime | new |
| `validate_jwt_with_resolver` — multi-issuer counterpart of `validate_jwt_generic` | new |
| `ValidationError::UntrustedIssuer` / `::MissingIssuer` | new variants |

`JwtStrategy` now holds a `Box<dyn JwksResolver>` instead of a `JwksCache` (private field; no API impact).

## Security: no fallback, ever

An `iss` outside a non-empty trust map is rejected with `UntrustedIssuer`; a token with no usable string `iss` is rejected with `MissingIssuer`. There is deliberately **no** fallback to `jwks_url` — a default JWKS for unknown issuers is exactly issuer confusion, and this is stated in the `JwksResolver` doc comment as a contract for third-party implementors.

Three defences, in order:

1. The resolver picks the JWKS from `iss`, with no default arm.
2. Every trusted issuer name is passed to `Validation::set_issuer` — which always took a slice; the config simply had no way to express more than one name (limitation 1 in the issue). So a token verified with B's key still has to *say* it came from B.
3. Multi-issuer mode adds `iss` to `required_spec_claims`, so even a lax custom resolver cannot let an `iss`-less token through.

The `iss` used to select a key is read **before** verification (`unverified_issuer`, a hand-rolled base64url payload read). That is safe because the same claim is re-checked after verification against the trusted set, and the payload is signature-covered — an attacker choosing `iss` only chooses whose key their signature must satisfy. The reasoning is written out on the function.

If `jwks_url` is set alongside a trust map, it is folded in as one more entry **only** when `issuer` names it. Without an `issuer` it is dropped with a `tracing::warn!` rather than becoming a default arm.

## Additive vs breaking

**Additive.** With an empty `trusted_issuers` map, `JwtStrategy::new` builds exactly the previous verifier: one JWKS, `iss` checked only if `issuer` was set, `iss` not required, same `Ok(None)` rejection semantics. Pinned by `single_issuer_configuration_is_unchanged_by_multi_issuer_support`, which also pins the *pre-existing* laxity that a token omitting `iss` entirely is still accepted in single-issuer mode (see "left open" #3).

All construction goes through `ValidationConfigBuilder` — no struct literal is added anywhere outside the defining crate — so this is compatible with #250 making `ValidationConfig` `#[non_exhaustive]`.

**Two things that are technically breaking under strict semver**, neither of which affects anything in-tree:

- `ValidationError` gains two variants. It is not `#[non_exhaustive]`, so a downstream exhaustive `match` would stop compiling. Making it `#[non_exhaustive]` is the same class of decision as #250 and is **not** done here.
- `ValidationConfig` gains a field, which is the exact hazard #247/#250 is about. Unavoidable for this feature; the builder covers it.

Nothing here touches `Cargo.toml`, the version, or `release-plz.toml`. The commit is a plain `feat`, which lands on the same `0.6.0` #250 already argues for.

## Expected conflict with #250

Both touch `crates/authkestra-resource/src/jwt.rs` and `tests/jwt_test.rs`. **A conflict is expected and the coordinator will resolve it.** This branch is deliberately *not* rebased onto #250.

The edit to the `ValidationConfig` struct definition itself is 6 added lines, appended after the last field, so #250's `#[non_exhaustive]` (added above the struct) should merge around it:

```rust
     pub require_cert_binding: bool,
+    /// Trust map of `iss` -> JWKS endpoint, for a resource server that accepts
+    /// tokens from several issuers. Empty by default, which keeps the
+    /// single-issuer `jwks_url` path unchanged. When non-empty, an `iss` absent
+    /// from this map is rejected outright and never falls back to `jwks_url`.
+    /// See [`IssuerTrustMap`] and issue #243.
+    pub trusted_issuers: BTreeMap<String, String>,
 }
```

The other in-file edits are away from that struct: the error enum, a new resolver section after `JwksCache`, the builder impl, `JwtStrategy` + two new free functions, and appended tests. Tests are appended at the end of `jwt_test.rs`, where #250 also appends one — a mechanical conflict.

## Left as a maintainer decision

1. **`jwks_url: String` vs `Option<String>`.** A trust-map-only config has no single endpoint, so `build()` leaves `jwks_url` empty (and an empty `jwks_url` is never fetched from — `build_resolver` skips it). `Option<String>` would model this honestly but changes a public field's type, which is breaking. Not done. `build()` still panics if neither a `jwks_url` nor any `trusted_issuer` was set.
2. **Should `ValidationError` become `#[non_exhaustive]`?** See above. Related to #247; not decided here.
3. **Should single-issuer mode also require `iss`?** Today, with `issuer` configured, `jsonwebtoken` only checks `iss` when the claim is present, so a token omitting it is accepted. That is pre-existing behaviour, unchanged here and now pinned by a test with a comment saying so. Multi-issuer mode does *not* inherit it. Tightening single-issuer mode would be a behavioural break for deployed verifiers — a maintainer call, not mine.
4. **`UntrustedIssuer` surfaces as `Err(AuthError::Token)` from `JwtStrategy::authenticate`, not `Ok(None)`**, matching how `MissingKid` already behaves (there is an existing test blessing that). Under `AuthPolicy::FirstSuccess` this aborts the strategy chain instead of falling through to, say, an API-key strategy. Only reachable when a trust map is configured, so no existing deployment changes. Flagging it in case `Ok(None)` is preferred.

## Verification

Ran on the toolchain CI resolves `stable` to (1.98.0), scoped to the touched crate:

| Command | Result |
| --- | --- |
| `cargo fmt --all -- --check` | pass |
| `cargo +1.98.0 clippy -p authkestra-resource --all-targets --all-features -- -D warnings` | pass |
| `cargo test -p authkestra-resource --all-features` | 9 unit + 19 integration, all pass |

No `#[allow]` was added anywhere.

### The multi-issuer tests are load-bearing (measured, not asserted)

**(a) Reverted the per-issuer resolution** — made `build_resolver` hand every issuer the same single JWKS, i.e. the pre-#243 shape:

```
test multi_issuer_verifies_each_token_against_its_own_issuers_jwks ... FAILED
test multi_issuer_rejects_a_token_whose_issuer_is_not_in_the_trust_map ... FAILED
test multi_issuer_never_falls_back_to_the_single_issuer_jwks_url ... FAILED

authenticate should not error for https://issuer-b.example: Token error: Key not found in JWKS
an untrusted issuer must surface as an error, not as an identity: None
expected an untrusted-issuer rejection, got: Token error: Key not found in JWKS
```

**(b) Introduced the actual vulnerability** — gave `IssuerTrustMap::resolve` a default arm returning the first cache for an unknown issuer:

```
test multi_issuer_rejects_a_token_whose_issuer_is_not_in_the_trust_map ... FAILED
test multi_issuer_never_falls_back_to_the_single_issuer_jwks_url ... FAILED

expected UntrustedIssuer(https://issuer-c.example), got Jwt(Error(InvalidIssuer))
expected an untrusted-issuer rejection, got: Token error: Key not found in JWKS
```

Worth reading precisely what (b) shows: with the fallback in place the token was *still* ultimately rejected — by layer 2, the accepted-issuer list. That is exactly the "rejected incidentally" failure mode, and the tests demand the specific `UntrustedIssuer` reason so they cannot pass on luck.

Both experiments were reverted; the final tree is the green one above.

### DIFF
--- a/crates/authkestra-oidc/src/error.rs
+++ b/crates/authkestra-oidc/src/error.rs
@@ -66,6 +66,18 @@ impl From<authkestra_resource::jwt::ValidationError> for OidcError {
             authkestra_resource::jwt::ValidationError::MissingKid => {
                 OidcError::ValidationError("Token is missing a required 'kid' header".to_string())
             }
+            // Multi-issuer resolution (#243). Both are validation failures from
+            // an RP's point of view: the token names an issuer this verifier
+            // holds no JWKS for, or names none at all so no JWKS can be picked.
+            authkestra_resource::jwt::ValidationError::UntrustedIssuer(issuer) => {
+                OidcError::ValidationError(format!(
+                    "Token issuer {issuer:?} is not in the configured trust map"
+                ))
+            }
+            authkestra_resource::jwt::ValidationError::MissingIssuer => OidcError::ValidationError(
+                "Token is missing a required 'iss' claim; an issuer is needed to select a JWKS"
+                    .to_string(),
+            ),
             authkestra_resource::jwt::ValidationError::Paseto(e) => OidcError::ValidationError(e),
             authkestra_resource::jwt::ValidationError::Validation(e) => {
                 OidcError::ValidationError(e)
--- a/crates/authkestra-resource/README.md
+++ b/crates/authkestra-resource/README.md
@@ -44,6 +44,30 @@ let config = ValidationConfig::builder()
 let strategy = JwtStrategy::new(config);
 ```
 
+### Trusting several issuers
+
+One verifier can accept tokens from several issuers, each with its own JWKS
+endpoint. The token's `iss` claim selects which JWKS verifies it:
+
+```rust
+use authkestra_resource::jwt::{JwtStrategy, ValidationConfig};
+
+let config = ValidationConfig::builder()
+    .trusted_issuer("https://tenant-a.example", "https://tenant-a.example/.well-known/jwks.json")
+    .trusted_issuer("https://tenant-b.example", "https://tenant-b.example/.well-known/jwks.json")
+    .audience("my-app")
+    .build();
+
+let strategy = JwtStrategy::new(config);
+```
+
+An `iss` that is not in the trust map is rejected: there is deliberately no
+fallback endpoint, because a default JWKS for unknown issuers is issuer
+confusion. A token with no `iss` at all is rejected too, since nothing names a
+key. For issuers that are only known at runtime, implement `JwksResolver` and
+pass it to `JwtStrategy::with_resolver` — the same rule applies, an
+implementation must reject issuers it does not recognise.
+
 ## Related Crates
 
 - `authkestra-engine`: Foundational types and the Engine orchestrator.
--- a/crates/authkestra-resource/src/jwt.rs
+++ b/crates/authkestra-resource/src/jwt.rs
@@ -458,7 +458,7 @@ impl<I> JwtStrategy<I> {
         let resolver = build_resolver(&config);
         Self {
             resolver,
-            validation: build_validation(&config),
+            validation: build_validation(&config, !config.trusted_issuers.is_empty()),
             require_cert_binding: config.require_cert_binding,
             _marker: std::marker::PhantomData,
         }
@@ -477,7 +477,9 @@ impl<I> JwtStrategy<I> {
     pub fn with_resolver(config: ValidationConfig, resolver: Box<dyn JwksResolver>) -> Self {
         Self {
             resolver,
-            validation: build_validation(&config),
+            // Unconditionally multi-issuer: supplying your own resolver *is* the
+            // multi-issuer case, whether or not a static trust map accompanies it.
+            validation: build_validation(&config, true),
             require_cert_binding: config.require_cert_binding,
             _marker: std::marker::PhantomData,
         }
@@ -532,7 +534,7 @@ fn build_resolver(config: &ValidationConfig) -> Box<dyn JwksResolver> {
 }
 
 /// Builds the claim-validation half of a [`JwtStrategy`] from a config.
-fn build_validation(config: &ValidationConfig) -> Validation {
+fn build_validation(config: &ValidationConfig, multi_issuer: bool) -> Validation {
     let mut validation = Validation::new(config.algorithms[0]);
     validation.algorithms = config.algorithms.clone();
 
@@ -552,11 +554,17 @@ fn build_validation(config: &ValidationConfig) -> Validation {
         validation.set_issuer(&accepted_issuers);
     }
 
-    if !config.trusted_issuers.is_empty() {
+    if multi_issuer {
         // In multi-issuer mode `iss` selects the key, so a token without one must
         // never reach the decode step. The resolver already refuses it; requiring
         // the claim here means a custom resolver that is laxer than it should be
         // still cannot let an `iss`-less token through.
+        //
+        // Gated on `multi_issuer` rather than on `trusted_issuers` being
+        // non-empty: a `with_resolver` caller resolves issuers dynamically and so
+        // configures no trust map at all, which is exactly the case this backstop
+        // was written for. Inferring it from the map left that path with neither
+        // `set_issuer` nor this check -- raised in review of #254.
         validation.required_spec_claims.insert("iss".to_string());
     }
 
--- a/crates/authkestra-resource/src/jwt.rs
+++ b/crates/authkestra-resource/src/jwt.rs
@@ -7,9 +7,12 @@ use authkestra_engine::{
         Claims,
     },
 };
+use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine as _};
 use http::request::Parts;
 use jsonwebtoken::{decode, decode_header, Algorithm, Validation};
 use serde::Deserialize;
+use std::collections::BTreeMap;
+use std::sync::Arc;
 use std::time::{Duration, Instant};
 use thiserror::Error;
 use tokio::sync::RwLock;
@@ -31,6 +34,12 @@ pub enum ValidationError {
         "Token is missing a 'kid' header, which is required by this JWKS cache's strict validation policy"
     )]
     MissingKid,
+    #[error("Token issuer {0:?} is not in the configured trust map; no JWKS is trusted for it")]
+    UntrustedIssuer(String),
+    #[error(
+        "Token carries no string 'iss' claim, which is required to resolve a signing key when more than one issuer is trusted"
+    )]
+    MissingIssuer,
     #[error("PASETO error: {0}")]
     Paseto(String),
     #[error("Discovery error: {0}")]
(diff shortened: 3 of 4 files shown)


stephane-segning:
## Adversarial review — issuer confusion & blast radius

**The issuer-confusion design is sound — the reviewer traced every branch and could not construct a cross-issuer bypass.** But it found a build-breaking blocker and one advertised defence that was absent exactly where advertised. Both fixed in `71cb3e2`.

---

### 🔴 BLOCKER — the PR did not compile in-tree (fixed)

`crates/authkestra-oidc/src/error.rs:48` matches `ValidationError` **exhaustively, with no wildcard arm**. Adding `UntrustedIssuer` and `MissingIssuer` broke it:

```
error[E0004]: non-exhaustive patterns: `ValidationError::UntrustedIssuer(_)`
              and `ValidationError::MissingIssuer` not covered
  --> crates/authkestra-oidc/src/error.rs:48:15
```

**CI confirms it independently** — `test`, `lint` and `coverage` were all red on this PR.

This directly refutes the PR body: *"Two things that are technically breaking under strict semver, **neither of which affects anything in-tree**."* It affects a sibling crate in the same workspace. The reason it was missed is that verification was scoped to `-p authkestra-resource`; the gate that catches this is `cargo check --workspace`, which I've now run.

Fixed by adding the two arms. It also converts your open question #2 (`#[non_exhaustive]` on `ValidationError`) from hypothetical into **already-answered-by-necessity** — though whether to *also* mark the enum `#[non_exhaustive]` for future additions remains your call.

### 🔴 HIGH — `required_spec_claims` was off on the `with_resolver` path (fixed)

PR body: *"Multi-issuer mode adds `iss` to `required_spec_claims`, so **even a lax custom resolver** cannot let an `iss`-less token through."*

`jwt.rs:555` gated that on `!config.trusted_issuers.is_empty()` — but a `with_resolver` caller resolves issuers *dynamically* and therefore configures no trust map at all. So neither `set_issuer` (also skipped when the accepted set is empty) nor `required_spec_claims` fired. **Both backstops were off precisely where the docs said they applied.** The in-code comment was self-refuting: it promised protection against a lax resolver on the one path that didn't get it.

Measured in review with a deliberately lax resolver:
```
PROBE7 with_resolver, no configured issuers, iss-less token -> Ok(Some(Ident { sub: "u" }))
```

`build_validation` now takes an explicit `multi_issuer: bool`; `with_resolver` passes `true` unconditionally, since supplying a resolver *is* the multi-issuer case. Pinned by `with_resolver_rejects_an_issuerless_token_even_with_no_trust_map`, whose resolver is maximally lax so the rejection must come from `Validation`. **Verified load-bearing** — restoring the inferred gate fails it with exactly the accepted-token output above.

Compounding and worth noting: `JwtStrategy::with_resolver`, `SingleJwksResolver`, and any third-party `impl JwksResolver` had **zero test coverage** — the tests only exercised the free `validate_jwt_with_resolver` with the built-in `IssuerTrustMap`.

---

### ✅ Verified sound — the core design

1. **No fallback arm anywhere.** Every branch read: `build_resolver` (`jwt.rs:500-531`) → empty map gives `SingleJwksResolver`, non-empty gives `IssuerTrustMap`; `IssuerTrustMap::resolve` (`jwt.rs:237-254`) has exactly `Some(cache)` / `None => Err(UntrustedIssuer)`. No `unwrap_or`, no default arm. The `_ => {}` at `jwt.rs:528` is inside `match &config.issuer` and only decides whether to fold in `jwks_url` — not a resolution path.
2. **No cross-issuer key confusion.** Selection is exact-match on the `iss` string, and layer 2 checks that same string. Selection key ≡ validated `iss` by construction.
3. **The two `iss` parses agree, or fail closed.** Duplicate `iss` members: `unverified_issuer` uses `serde_json::Value` (last-wins) while `ClaimsForValidation` derives `Deserialize` (duplicate-field error) — both orderings measured, neither authenticates. Array `iss` → `None` → `Err(MissingIssuer)` before decode.
4. **`kid` collisions handled** — lookup is in the *resolved* cache only, and the existing `..._claiming_issuer_a_but_signed_with_issuer_bs_key` test uses a genuinely shared `"shared-kid"`.
5. **No background-task leak — the premise was false.** `JwksCache` spawns nothing; refresh is lazy inside `get_jwks`/`get_key`. `grep` for `tokio::spawn|JoinHandle` in the crate returns nothing.
6. **Single-issuer regression: none.** "Byte-for-byte" slightly overstates it (extra parse, different internal error variant) but it's behaviourally exact at the API surface.

**Your experiment (b) reproduced byte-identically**, and its reading is right and important: with the fallback in, the token was *still rejected* — by layer 2 — so the tests only catch it because they assert the specific reason. They cannot pass on luck. It actually caught one more than reported (`trust_map_rejects_unknown_and_missing_issuers_without_falling_back` fails too, aborting before the integration tests).

---

### 📋 Remaining, not actioned

- **M3 — `UntrustedIssuer`/`MissingIssuer` produce a remotely-triggerable HTTP 500.** `authkestra-axum/src/lib.rs:200-203` maps these to `Internal` → `INTERNAL_SERVER_ERROR`; same in actix. Any unauthenticated request with a valid-looking JWT and unknown `iss` guarantees a 5xx — scriptable into 5xx-alert flooding and error-log spam at `tracing::error!`. **Mitigating:** `#[error("... {0:?}")]` uses Debug formatting, so `\r\n` is escaped — no header injection (measured), and axum serves `String` as `text/plain`, so no XSS. That `{0:?}` is a good call and should stay. This is your open question #4, but decide it knowing it's a 500, not just a `FirstSuccess` chain abort.
- **L4 — `unverified_issuer` accepts >3 segments and reads the wrong one.** `jwt.rs:700-709` doesn't check for exactly three segments; for `h.p1.p2.s` it reads `p1` while `jsonwebtoken` reads `p2`. **The two parses disagree by construction** — the classic select-one-key/validate-another setup. Currently fail-closed only *incidentally* (`decode_header` gets `"h.p1"`, which contains a `.`, which base64 rejects). Cheap fix: reject a 4th segment explicitly.
- **L5 — layer 2's accepted set is a strict superset of layer 1's resolvable set.** `build_resolver` folds in `config.issuer` only when `jwks_url` is non-empty; `build_validation` adds it unconditionally. Fail-closed today (measured), but the backstop is widened past the thing it backstops.
- **L6 — `Jwt<T>`, the most obvious extractor, cannot reach multi-issuer mode** (bound to `Arc<JwksCache>`, calls `validate_jwt_generic` directly). Worth a README line.
- **L7/L8** — silent `or_insert_with` drop when `issuer` is also a `trusted_issuer` key; `unverified_issuer` parsing on every single-issuer request where the result is discarded.

### 🔎 New evidence on your open question #3

`set_issuer` is bypassed by a **non-string** `iss`, not just an absent one — `"iss": 1234` deserializes to `TryParse::FailedToParse`, hitting `_ => {}` (`jsonwebtoken validation.rs:301-311`). Measured on this branch in single-issuer mode with `.issuer(ISS_A)` configured: `Ok(Some(Ident { sub: "u" }))`. **`main` behaves identically — this PR does not introduce it.** But it's strictly worse than the absent-`iss` laxity you pinned, and the same one-line fix covers both. Multi-issuer mode is already safe (measured `Err(MissingIssuer)`).

The #250 conflict was explicitly **not** treated as a finding — the 6-line appended struct edit is as mechanical as you described.

<sub>🤖 Adversarial review by an Opus agent. Every branch traced; experiment (b) reproduced; the blocker independently confirmed by CI; both fixes verified load-bearing before posting.</sub>

marcjazz (approved):

stephane-segning:
Review findings addressed in `1336eab`, which also merges the now-landed #250.

**Merge with #250** — resolved as predicted. #250 marked `ValidationConfig` `#[non_exhaustive]` and deleted a stray doc line sitting above the struct; this branch added the resolver section just before it. Kept the resolver and **honoured #250's deletion** rather than resurrecting the line. Git had also interleaved the two branches' appended tests, so both suites are preserved — this branch's multi-issuer tests plus #250's `builder_defaults_match_documented_behaviour`. 21 integration tests pass.

**L4 — parser disagreement (fixed).** `unverified_issuer` accepted more than three segments and read the wrong one: for `h.p1.p2.s` it returned `p1` while `jsonwebtoken` reads `p2`, so the issuer used to *select* a key and the claims that get *validated* came from different segments. Fail-closed only by accident (`decode_header` got `"h.p1"`, whose `.` base64 rejects) — not a property worth resting on. Now rejected explicitly.

**L5 — backstop widened past what it backstops (fixed).** `build_validation` folded `config.issuer` in unconditionally while `build_resolver` folds it only when `jwks_url` is non-empty, so `.issuer(A).trusted_issuer(B, urlB)` with no `jwks_url` left `Validation` accepting `{A, B}` while the resolver knew only `{B}`. Both now gate on the same condition.

`fmt` clean, `clippy +1.98.0 --workspace` exits 0, `cargo check --workspace` clean, 9 unit + 21 integration pass. The branch merges cleanly against current `main`.

**Still yours to decide** (unchanged from the review):
- **M3** — `UntrustedIssuer`/`MissingIssuer` surface as a remotely-triggerable **HTTP 500**, not just a `FirstSuccess` chain abort. `Ok(None)` would be safer. Mitigating: `{0:?}` Debug formatting escapes CRLF, so no header injection.
- Whether `ValidationError` should be `#[non_exhaustive]` (I added the two arms to `authkestra-oidc` to unbreak the build; that is orthogonal to marking the enum).
- Whether single-issuer mode should also require `iss` — note the review found a **non-string** `iss` (`"iss": 1234`) also bypasses `set_issuer`, which is pre-existing on `main` and strictly worse than the absent-`iss` laxity. Same one-line fix covers both.

Issue #194, opened 2026-08-13
[Examples] Migrate Axum and Actix examples to the unified engine architecture


marcjazz:
Sub-task of Epic #192

**Context:**
Examples in `crates/authkestra/examples` are outdated. We need clear 'Golden Path' examples as defined in RFC-001.

> [!WARNING]
> **Dependency Graph Complexity**
> Be highly self-aware before moving examples out of the `authkestra` facade crate. The examples reside in the facade crate for a structural reason: it sits at the top of the dependency graph. 
> 
> Moving framework-specific examples (like `axum_...` or `actix_...`) down into `authkestra-engine` would introduce circular dependencies (e.g., `authkestra-axum` as a dev-dependency in `authkestra-engine`). Even moving them into adapter crates like `authkestra-axum` might be tricky if the examples require other leaf crates (like specific session stores or providers). 
> 
> **Recommendation:** If it is too complex to move them without breaking the dependency graph, the examples should remain in `crates/authkestra/examples`. In that case, the task should solely focus on **refactoring** the existing examples in place to use the new `authkestra-engine` and `Engine::builder()` patterns, rather than relocating them.

**Tasks:**
- [ ] Refactor `axum_basic_setup.rs` and `axum_oauth_stateless.rs` to use `authkestra-engine` and `authkestra-axum`.
- [ ] Refactor corresponding Actix examples.
- [ ] Evaluate if it's safe to move examples out of `crates/authkestra/examples` without causing circular dependencies. If not, keep them there and only update their code.

marcjazz:
> [!WARNING]
> **Dependency Graph Complexity**
> Be highly self-aware before moving examples out of the  facade crate. The examples reside in the facade crate for a structural reason: it sits at the top of the dependency graph. 
> 
> Moving framework-specific examples (like  or ) down into  would introduce circular dependencies (e.g.,  (dev-dep) ->  -> ). Even moving them into adapter crates like  might be tricky if the examples require other leaf crates (like specific session stores or providers). 
> 
> **Recommendation:** If it's too complex to move them without breaking the dependency graph, the examples should remain in . In that case, the task should solely focus on **refactoring** the existing examples in place to use the new  and  patterns, rather than relocating them.

PR #255, opened 2026-08-27
refactor(examples): modernise examples for the unified engine (#194)

stephane-segning:
Refactors the examples in `crates/authkestra/examples` **in place** around the `Engine::builder()` / unified-engine surface (RFC-001). Nothing was relocated — see the dependency-graph section for the evidence.

## Dependency graph: examples stay in the facade crate

I did not take the issue's `[!WARNING]` on faith; I reproduced the failure. Adding `authkestra-axum` as a dev-dependency of `authkestra-engine`:

- **builds fine locally** — `cargo check -p authkestra-engine --all-features` succeeds, because cargo tolerates dev-dependency cycles, and `cargo tree` shows the cycle plainly (`authkestra-engine [dev-dependencies] -> authkestra-axum v0.5.5 -> authkestra-engine v0.5.5 (*)`);
- **breaks the release**, which is the part that matters. Simulating a release-plz version bump to 0.5.6 and running `cargo publish --dry-run -p authkestra-engine`:

```
error: failed to prepare local package for uploading

Caused by:
  failed to select a version for the requirement `authkestra-axum = "^0.5.6"`
  candidate versions found which didn't match: 0.5.5, 0.5.4, 0.5.3, ...
  location searched: crates.io index
  required by package `authkestra-engine v0.5.6`
```

`authkestra-engine` cannot be published before `authkestra-axum`, which cannot be published before `authkestra-engine`. This is exactly the deadlock that 7f9ea0d ("fix: move examples to root crate to permanently resolve cyclic publish failures") already fixed by moving these examples *into* the facade. Relocating them would re-introduce a solved bug, so **nothing was moved**. The experiment was reverted; no `Cargo.toml` outside `crates/authkestra` is touched by this PR.

## Per-example status

| Example | Status |
| --- | --- |
| `axum_basic_setup.rs` | **Migrated** — canonical imports, tracing, `PORT`, 401 on no session, `/api/providers` |
| `axum_oauth2_github.rs` | **Migrated** — as above; **fixed** wrong default `redirect_uri` |
| `axum_oidc_google.rs` | **Migrated** — as above; **fixed** wrong default `redirect_uri` |
| `axum_oauth_stateless.rs` | **Migrated** — as above; **fixed** advertised login path and an `unwrap()` panic |
| `actix_basic_setup.rs` | **Migrated** — Actix counterpart, kept deliberately parallel to the Axum file |
| `actix_oauth2_github.rs` | **Migrated** — as above; **fixed** wrong default `redirect_uri` |
| `actix_oidc_google.rs` | **Migrated** — as above; **fixed** wrong default `redirect_uri` |
| `actix_oauth_stateless.rs` | **Migrated** — as above; **fixed** advertised login path and an `unwrap()` panic |
| `axum_session_redis.rs` | **Migrated** — reframed as "basic_setup with one line changed" |
| `axum_sql_store.rs` | **Migrated** — same; documented *why* the pre-existing `#![allow(deprecated)]` is legitimate |
| `axum_data_layer_macros.rs` | **Migrated** — same; added `/api/user` so the bundled page works |
| `axum_mfa_server.rs` | **Migrated** — gained a module doc (it had none), `with_webauthn` shorthand, tracing |
| `axum_op_server.rs` | **Migrated** — **fixed** `issuer`/bind-port mismatch and the literal `"issuer"` `iss` claim |
| `actix_op_server.rs` | **Migrated** — same fixes |
| `axum_op_server_sqlx.rs` | **Migrated** — same fixes |
| `actix_op_server_sqlx.rs` | **Migrated** — same fixes |
| `axum_op_server_custom_grant.rs` | **Migrated** — same fixes |
| `actix_op_server_custom_grant.rs` | **Migrated** — same fixes |
| `axum_resource_server.rs` | **Migrated** — module doc, tracing; now points at the `Guard` example as preferred |
| `axum_resource_server_strategy.rs` | **Migrated** — module doc, tracing, `PORT` |
| `actix_resource_server_strategy.rs` | **Migrated** — module doc, tracing, `PORT` |
| `axum_op_server_attestation.rs` | **Left as-is** — already on current patterns, already binds port 0, already has a thorough module doc. |
| `actix_op_server_attestation.rs` | **Left as-is** — out of scope for this PR. *(Correction, raised in review: it does **not** already bind port 0 — it hardcodes `127.0.0.1:8092`. Both attestation examples also still use `println!` rather than `tracing`, so "nothing to migrate" was wrong for this pair.)* |
| `axum_devsig/` | **Left as-is** — out of scope for this PR. *(Correction, raised in review: the stated reason — that editing these would collide with #242 — does **not** hold. PR #253 touches only `Cargo.lock` and eight files under `crates/authkestra-devsig/`, none under `examples/`. Deferring is still fine; there is no dependency to wait on.)* |
| `actix_devsig/` | **Left as-is** — same, and it hardcodes `127.0.0.1:8090`. These two are the least modernised examples in the tree. |

**Nothing deleted.**

## Bugs found and fixed

These were not cosmetic — an example that documents a wrong URL is worse than an outdated one:

1. `axum/actix_oauth2_github` and `axum/actix_oidc_google` defaulted `redirect_uri` to `/auth/{provider}/callback`. The adapters wire `/auth/callback/{provider}`. Copying these examples gave you a callback that 404s.
2. The stateless examples printed `/auth/github` as the login URL; the route is `/auth/login/github`.
3. `examples/static/index.html` linked to `/auth/login`, which is not a route at all. It now queries a new `/api/providers` endpoint and links `/auth/login/{provider}` per configured provider (and says so honestly when an example configures none).
4. The Axum session examples returned **200** with an `error` body when unauthenticated, so `res.ok` was true and the static page rendered "Welcome, undefined". Now 401, matching the Actix side.
5. Every OP example set `OpConfig::issuer` to `http://localhost:3000` (or a literal 8080) while binding elsewhere, **and** signed tokens with the literal `iss` claim `"issuer"`. Both now derive from the bind port, so discovery and the id_token `iss` agree.
6. The stateless examples called `claims.identity.as_ref().unwrap()` — a panic on any validated token without an identity (e.g. client credentials). Now a 403.

## Ports and the `AddrInUse` landmine

- Every example reads `PORT` (default 3000; 8080 for OP servers) and derives its default redirect URI from it.
- `examples_e2e.rs` reserves a free port from the OS (`127.0.0.1:0`) and passes it via `PORT`, instead of assuming 3000 is free.
- `examples_e2e.rs` now kills the spawned example from a **`Drop` guard**. This is the actual orphan source: a failed assertion unwound past the manual `kill()`, leaving the example and its `cargo run` parent holding the port, so the *next* run failed with `AddrInUse` for reasons that looked nothing like the original failure.

The e2e test still asserts exactly what it asserted before (redirect status + redirect target is the wiremock URI). No test was weakened and no `#[allow]` was added.

## Verification

```
cargo fmt --all -- --check                                                  # clean
cargo +1.98.0 clippy -p authkestra --all-targets --all-features -- -D warnings  # clean
cargo build -p authkestra --examples --all-features                         # 25/25 build
cargo test -p authkestra --all-features -- --test-threads=2                 # 4 passed, 0 failed
```

`examples_e2e` passes in 18s, with the log showing each example on an OS-assigned port (40057, 39633, 40055, 44715, ...) rather than 3000.

Refs #194. Not auto-closing: four examples are intentionally left unmigrated (two already current, two owned by concurrent work).

---

## Corrections and review fixes

Two claims in the table above were wrong; both are corrected inline, and both were caught by adversarial review rather than by me.

Findings fixed since this PR was opened:

- **`5d67ada`** — the callback-path fix this PR is *named for* had no regression coverage. Reverting it left the e2e suite green, because the assertion only checked the redirect's prefix. The authorization URL already carries `redirect_uri`, so pinning it cost one assert.
- **`c916864`** — `index.html` collapsed "endpoint absent" and "no providers registered" into the same empty list, so it asserted "This example registers no OAuth providers" on three examples that never mount `/api/providers`. True for all three, but only by luck — the page cannot know it. It now distinguishes the two.
- **`c916864`** — the OP examples registered `redirect_uris` of `http://localhost:3000/callback`, a path no sibling example serves. Same defect class as the bug this PR exists to fix, left unfixed in the OP half. They now point at `/auth/callback/github`, so an OP on :8080 and an RP on :3000 compose without editing either.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -464,6 +464,7 @@ dependencies = [
  "tower-http",
  "tower-layer",
  "tower-service",
+ "tracing",
  "tracing-subscriber",
  "url",
  "uuid",
--- a/crates/authkestra/Cargo.toml
+++ b/crates/authkestra/Cargo.toml
@@ -44,6 +44,7 @@ p256 = { version = "0.13", features = ["pem", "ecdsa", "jwk"] }
 rand_core = { version = "0.6", features = ["getrandom"] }
 rand = "0.8"
 sha2 = "0.10"
+tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 async-trait = "0.1.91"
 webauthn-rs = "0.5.0"
--- a/crates/authkestra/examples/actix_basic_setup.rs
+++ b/crates/authkestra/examples/actix_basic_setup.rs
@@ -1,71 +1,118 @@
-//! # Actix Basic Setup Example
+//! # Actix Basic Setup — the smallest useful `Engine`
 //!
-//! This example demonstrates the most basic setup of Engine with Actix.
-//! It uses an in-memory session store.
+//! The Actix counterpart of `axum_basic_setup.rs`. The engine construction is
+//! byte-for-byte identical; only the adapter call differs
+//! ([`ActixExt::actix_scope`] instead of `axum_router`), which is the whole
+//! point of the framework-agnostic core.
+//!
+//! ```sh
+//! cargo run -p authkestra --example actix_basic_setup --all-features
+//! ```
+//!
+//! Set `PORT` to bind somewhere other than 3000.
 
 use actix_files::Files;
 use actix_web::{get, web, App, HttpResponse, HttpServer, Responder};
 use authkestra_actix::{ActixExt, ActixState, AuthSession};
-use authkestra_engine::auth::SessionStore;
-use authkestra_engine::flow::Engine;
-use authkestra_engine::SessionConfig;
+use authkestra_engine::store::memory::MemoryStore;
+use authkestra_engine::{AkWebAppEngine, Engine, SessionConfig, SessionStore};
 use serde_json::json;
 use std::sync::Arc;
 
+/// `AkWebAppEngine` is the alias for a session-configured engine. Using the
+/// alias (rather than spelling out the typestate generics) keeps the compile
+/// error legible when a required builder call is missing.
 #[derive(Clone, ActixState)]
 struct AppState {
     #[authkestra(engine)]
-    auth: authkestra_engine::AkWebAppEngine,
+    auth: AkWebAppEngine,
 }
 
 #[actix_web::main]
 async fn main() -> std::io::Result<()> {
-    // Session Store
-    let session_store: Arc<dyn SessionStore> =
-        Arc::new(authkestra_engine::store::memory::MemoryStore::default());
+    // `RUST_LOG=authkestra=debug` surfaces the engine's own instrumentation.
+    tracing_subscriber::fmt()
+        .with_env_filter(
+            tracing_subscriber::EnvFilter::try_from_default_env()
+                .unwrap_or_else(|_| "info,authkestra=debug".into()),
+        )
+        .init();
 
-    let authkestra = Engine::builder()
+    // 1. Pick a storage backend. `SessionStore` is a trait, so swapping
+    //    `MemoryStore` for `RedisStore` or `SqlKvStore` is a one-line change.
+    let session_store: Arc<dyn SessionStore> = Arc::new(MemoryStore::default());
+
+    // 2. Build the engine. The typestate builder only exposes session APIs
+    //    once `session_store` has been supplied.
+    let engine = Engine::builder()
         .session_store(session_store)
         .session_config(SessionConfig {
-            secure: false, // For local development
+            secure: false, // plain HTTP for local development
             ..Default::default()
         })
         .build();
 
-    let state = AppState {
-        auth: authkestra.clone(),
-    };
+    let state = AppState { auth: engine };
 
-    println!("🚀 Actix Basic Setup running on http://localhost:3000");
+    let port = port_from_env();
+    tracing::info!(%port, "Actix basic setup listening on http://localhost:{port}");
 
     HttpServer::new(move || {
         let app_state = state.clone();
         let config_state = app_state.clone();
         App::new()
             .app_data(web::Data::new(app_state.clone()))
+            // `configure_authkestra` (from `ActixState`) registers the session
+            // store and config as app data for the extractors to find.
             .configure(move |cfg| config_state.configure_authkestra(cfg))
             .service(get_user)
+            .service(list_providers)
+            // 3. Mount the engine's `/auth/*` scope.
             .service(app_state.auth.actix_scope())
             .service(
                 Files::new("/", concat!(env!("CARGO_MANIFEST_DIR"), "/examples/static"))
                     .index_file("index.html"),
             )
     })
-    .bind(("0.0.0.0", 3000))?
+    .bind(("0.0.0.0", port))?
     .run()
     .await
 }
 
-/// API endpoint to get current user info from session
+/// Returns the session identity, or 401 when there is no valid session.
 #[get("/api/user")]
 async fn get_user(session: Option<AuthSession>) -> impl Responder {
     match session {
-        Some(AuthSession(session)) => HttpResponse::Ok().json(json!({
-            "id": session.identity.external_id,
-            "username": session.identity.username,
-            "email": session.identity.email,
-            "provider": session.identity.provider_id,
-        })),
-        None => HttpResponse::Unauthorized().json(json!({ "error": "Not authenticated" })),
+        Some(AuthSession(session)) => {
+            tracing::debug!(user = %session.identity.external_id, "session resolved");
+            HttpResponse::Ok().json(json!({
+                "id": session.identity.external_id,
+                "username": session.identity.username,
+                "email": session.identity.email,
+                "provider": session.identity.provider_id,
+            }))
+        }
+        None => {
+            tracing::debug!("no valid session on request");
+            HttpResponse::Unauthorized().json(json!({ "error": "Not authenticated" }))
+        }
     }
+}
+
+/// Lists the OAuth providers registered on the engine so the bundled static
+/// page can render one login button per provider. Empty for this example.
+#[get("/api/providers")]
+async fn list_providers(state: web::Data<AppState>) -> impl Responder {
+    let providers: Vec<&String> = state.auth.providers.keys().collect();
+    HttpResponse::Ok().json(json!({ "providers": providers }))
+}
+
+/// Bind port, overridable via `PORT` so several examples (and the e2e test
+/// harness, which passes a port it has already confirmed to be free) can run
+/// without colliding on 3000.
+fn port_from_env() -> u16 {
+    std::env::var("PORT")
+        .ok()
+        .and_then(|p| p.parse().ok())
+        .unwrap_or(3000)
 }
--- a/crates/authkestra/examples/actix_oauth2_github.rs
+++ b/crates/authkestra/examples/actix_oauth2_github.rs
@@ -1,17 +1,23 @@
-//! # Actix GitHub OAuth2 Example
+//! # Actix + GitHub OAuth2 (stateful sessions)
 //!
-//! This example demonstrates how to set up Engine with Actix for GitHub OAuth2 login.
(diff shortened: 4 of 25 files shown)


stephane-segning:
## Adversarial review — factual accuracy of the examples migration

Examples are documentation: a wrong example teaches a wrong pattern to every reader. So the lens here was **are the PR's factual claims true**, verified by building and running rather than reading.

**Verdict: the claims check out.** All six were confirmed empirically, including the one that would have been most damaging if backwards. One Medium finding (fixed), two body-wording corrections needed.

---

### ✅ Claims verified

**1. "All 25 examples build."** True — `cargo build -p authkestra --examples --all-features` exits 0, 25 distinct binaries. `fmt` clean, `clippy +1.98.0 --all-targets --all-features -D warnings` exits 0 with **zero** warnings.

**2. The `redirect_uri` fix is in the CORRECT direction.** This was the claim most worth checking — if reversed, the PR would *break* working examples. The adapters wire callback-then-provider:
- `authkestra-axum/src/lib.rs:245` — `"/auth/callback/{provider}"` (and `:273` stateless)
- `authkestra-actix/src/lib.rs:56,63` — `scope("/auth")` + `"/callback/{provider}"`

And the old examples were genuinely wrong (`main:axum_oauth2_github.rs:56` — `".../auth/github/callback"`, same in three siblings). Confirmed correct. The table also doesn't over-claim: the two `*_oauth_stateless` examples already had the right path.

**3. `static/index.html` linked to a non-route.** True — `main` had `<a href="/auth/login">`; only `/auth/login/{provider}` exists.

**4. The `AddrInUse` root cause and `Drop` guard.** Old shape confirmed: `child.kill()` sat *after* both asserts, so a failing assertion unwound past it. The new guard is `let _guard = ChildGuard(child);` — a real named binding, not `let _ =`, so it survives to scope end.

Verified **empirically**, not by reading: breaking an example's authorize URL made the test fail, and `ps` afterwards showed **zero surviving example processes**. The guard fires on unwind.

**5. "Six real bugs."** All six are genuine defects, none a style change in disguise. The most severe is one the PR undersells — `main:axum_op_server.rs:39` passed `Some("issuer".to_string())`, which `TokenManager::new` puts straight into the `iss` claim, so the example OP minted `iss: "issuer"` and every conforming RP would reject it. Also real: a handler returning **200** with `{"error": "Not authenticated"}` (so the page rendered "Welcome, undefined"), and a remote-triggerable `.unwrap()` in a copied handler.

**6. `tracing` is dev-only and used.** Inside `[dev-dependencies]`, single `Cargo.lock` line, used throughout.

**AGENTS.md**, clarity and scope all assessed positively. The examples now *teach* the typestate builder rather than just using it. `port_from_env()` is duplicated across ~20 files, which is the right trade for files that must read standalone. The broader-than-asked scope is coherent — the issue's four named examples were already migrated; the rot was in the neighbours.

---

### 🔧 Medium — the headline fix had zero regression coverage (fixed in `5d67ada`)

Demonstrated, not theorised: reverting `axum_oauth2_github.rs` to the old wrong path left the suite **green**.

```
test test_all_oauth_examples_sequentially ... ok
```

The assertion only checked the redirect's *prefix* (that it points at wiremock). But the authorization URL already carries the answer — `...&redirect_uri=http%3A%2F%2Flocalhost%3A36049%2Fauth%2Fcallback%2Fgithub&...` — so pinning it costs one assert. Given this PR argues a wrong example is worse than an outdated one, the fix it is named for deserved a test.

**Not a weakening of anything**: the gap is pre-existing and identical on `main`; this PR just had the best opportunity to close it.

---

### 📋 Two body claims that don't survive checking

**"`actix_op_server_attestation.rs` — left as-is, already binds port 0."** False. That file (unchanged by this PR) has:
```rust
// crates/authkestra/examples/actix_op_server_attestation.rs:106-110
let addr: SocketAddr = "127.0.0.1:8092".parse().unwrap();
```
Only the *axum* attestation example binds `127.0.0.1:0`. Both attestation examples also still use `println!` (6 each) — the only non-devsig examples that do — so "already on current patterns" doesn't hold against this PR's own tracing bar either.

**"The devsig examples would collide with #242."** Doesn't hold. PR #253 touches `Cargo.lock` and eight files under `crates/authkestra-devsig/` — **zero under `crates/authkestra/examples/`**. Its public additions (`VerifyFailure`, `verify_jws_signature`, and test-only helpers) are not used by the examples, which call `DevSig::builder()` and `authkestra_devsig::ReplayStore`, both untouched.

Meanwhile the devsig examples are the *least* modernised in the tree (8 `println!` each, no `PORT`, hardcoded `127.0.0.1:8090`). Deferring them is entirely fine — but the honest reason is "out of scope for this PR", not "would collide with #253". Worth correcting so the next person doesn't wait on a phantom dependency.

### 📋 Follow-up material, not blocking

- **`/api/providers` missing on three static-serving examples** (`axum_session_redis`, `axum_sql_store`, `axum_data_layer_macros`). They 404 → the page renders "This example registers no OAuth providers", which is *true* but only by luck — the page can't distinguish "none configured" from "endpoint absent". `axum_basic_setup` registers the route despite also having none, which makes the omission look accidental.
- **The OP examples' registered client points at `http://localhost:3000/callback`**, which no sibling example serves (RPs serve `/auth/callback/{provider}`). Since the OP now defaults to 8080 specifically to run beside an RP on 3000, pointing it at the real path would make the two composable. Same defect class as bug #1 in the PR's own list.
- `.unwrap()`/`.expect()` inconsistency in OP setup paths; `axum_mfa_server.rs` names `with_totp` in a comment then visibly doesn't call it; the axum-only `axum_resource_server.rs` asymmetry isn't mentioned in the body.

**On `examples_e2e` not being weakened:** assertions and login paths are byte-identical to `main`, it caught a deliberately broken example, and it's arguably now *stricter* — an example ignoring `PORT` and binding 3000 would time out. The `free_port()` TOCTOU is named honestly in a doc comment rather than papered over, which is the right call for a teaching repo.

<sub>🤖 Adversarial review by an Opus agent. All 25 examples built, the adapter routes read directly, the `Drop` guard verified by inducing a panic and checking for orphans, and the coverage gap demonstrated by reverting the fix.</sub>

marcjazz (approved):

stephane-segning:
Review findings addressed. Two claims in the PR body were factually wrong and are now corrected inline there, marked as corrections rather than quietly edited.

**Fixed:**
- **`5d67ada`** — the callback-path fix this PR is *named for* had no regression coverage. Reverting it left the e2e suite green, since the assertion only checked the redirect's prefix. The authorization URL already carries `redirect_uri`, so pinning it cost one assert.
- **`c916864`** — `index.html` collapsed "endpoint absent" and "no providers registered" into the same empty list, so it asserted *"This example registers no OAuth providers"* on the three examples that serve the page without mounting `/api/providers`. True for all three, but only by luck — the page cannot know it. Now distinguishes the two. One file, rather than adding the route to three examples.
- **`c916864`** — the OP examples registered `redirect_uris` of `http://localhost:3000/callback`, which no sibling example serves. Same defect class as the bug this PR exists to fix, left unfixed in the OP half. Now `/auth/callback/github`, so an OP on :8080 and an RP on :3000 compose without editing either.

**Corrected in the body:**
- `actix_op_server_attestation.rs` does **not** already bind port 0 — it hardcodes `127.0.0.1:8092`. Both attestation examples also still use `println!`, so "nothing to migrate" was wrong for that pair.
- The devsig skip reason doesn't hold: #253 touches only `Cargo.lock` and eight files under `crates/authkestra-devsig/`, **none** under `examples/`. Deferring is fine; there was no collision to avoid.

25 examples build, `fmt` clean, `clippy +1.98.0 --all-targets` exits 0, `examples_e2e` + typestate tests pass. Merges cleanly against current `main`.

**Left as follow-ups**, not bundled into an already-broad PR: the `.unwrap()`/`.expect()` inconsistency in OP setup paths, the `with_totp` comment ordering in `axum_mfa_server.rs`, and the missing Actix counterpart to `axum_resource_server.rs`.

Issue #256, opened 2026-08-27
authkestra-op: device enrolment accepts low-order Ed25519 keys — OP mints attestations for keys nobody holds


stephane-segning:
## Summary

`authkestra-op`'s device-enrolment proof-of-possession verifies a challenge signature against a JWK that is **unauthenticated request input**, using the same non-strict EdDSA verifier reported in #242. A client can enrol a low-order Ed25519 public key — one for which no private key exists — and have the OP mint a real attestation binding it.

This is the same defect class as #242 but a **distinct, live path**: #242 concerns verification in `authkestra-devsig`, this concerns *issuance* in `authkestra-op`. #253 fixes the former and explicitly does not close #242 for that reason.

## Why this is not theoretical

Unlike the `authkestra-resource` JWKS path (where the key comes from a trusted OP), here the attacker supplies the key directly:

- `crates/authkestra-op/src/handlers/enrolment.rs:39` — `EnrolStartRequest { pub public_jwk: Value }` — attacker-controlled.
- `crates/authkestra-op/src/handlers/enrolment.rs:300-302`:
  ```rust
  let jwk = parse_public_jwk(&challenge.public_jwk)?;
  let signed_challenge = verify_challenge_signature(&req.challenge_signature, &jwk)?;
  ```
- `crates/authkestra-op/src/attestation.rs:263-290` — `parse_public_jwk` rejects private/symmetric members and checks `kty` only. **No small-order check.**
- `crates/authkestra-op/src/attestation.rs:373` — `AlgorithmParameters::OctetKeyPair(_) => Ok(Algorithm::EdDSA)`, so any OKP routes to the EdDSA verifier.
- `crates/authkestra-op/src/attestation.rs:429` — `jsonwebtoken::decode::<ChallengePayload>(...)`, i.e. `jsonwebtoken`'s non-strict `rust_crypto` backend.

## Attack

1. `POST` enrolment start with the identity point:
   ```json
   {"kty":"OKP","crv":"Ed25519","x":"AQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"}
   ```
2. Complete the challenge with the canned signature `(R = identity, S = 0)`. Under non-strict verification `[S]B - [k]A == R` holds for **every** message — both sides collapse to the identity for any challenge scalar.
3. The OP accepts the proof of possession and issues an attestation binding that `cnf.jkt`.

[...]

PR #265, opened 2026-08-27
fix(op): verify EdDSA strictly and refuse low-order enrolment keys (#256)

stephane-segning:
Refs #256.

## What I found, and how it differs from the issue

The issue was filed against `2731e04`. Since then #261 landed a low-order check in `parse_public_jwk`, which **does** close the literal vector in #256 (`crv: "Ed25519"` + identity point). I confirmed that before changing anything.

The defect itself was still live, via two gaps found while writing the reproduction:

**1. The `crv` gate is bypassable.** #261 wrote the check as `if params.curve == Ed25519 { ... }`. But `jsonwebtoken`'s `OctetKeyPairParameters::curve` is the *shared* `EllipticCurve` enum, so `{"kty":"OKP","crv":"P-256"}` deserialises into the `OctetKeyPair` variant and skips the gate — while `expected_algorithm` maps **every** `OctetKeyPair` to `Algorithm::EdDSA` and `DecodingKey::from_jwk` sends **every** `OctetKeyPair` to `from_ed_components`, both regardless of `crv`. So `x` is interpreted as Ed25519 bytes either way, and the key reaches the Ed25519 verifier having skipped the low-order check.

**2. The ceremony was saved by an accident, not a control.** Those keys did not actually mint an attestation on `main` — but only because `compute_cnf_jkt` fails afterwards with `InvalidKeyFormat`, since `jsonwebtoken` won't thumbprint a non-Ed25519 OKP key. That is an unrelated downstream step whose doc comment makes no security claim, and it would evaporate the moment thumbprinting grew support for another OKP curve.

Measured on `main` at `876927e`, the proof-of-possession check itself was fully bypassed:

```
PROBE[Ed25519] parse_public_jwk REJECTED: Ed25519 public key is a low-order point; ...
PROBE[P-256] parse_public_jwk ACCEPTED
PROBE[P-256]   verify_challenge_signature -> Ok("CHAL")      <-- forgery verified
PROBE[P-256]   compute_cnf_jkt -> Err(BadJwk("jwk thumbprint failed: InvalidKeyFormat"))
PROBE[P-384] parse_public_jwk ACCEPTED
PROBE[P-384]   verify_challenge_signature -> Ok("CHAL")      <-- forgery verified
PROBE[P-521] parse_public_jwk ACCEPTED
PROBE[P-521]   verify_challenge_signature -> Ok("CHAL")      <-- forgery verified
```

And with `parse_public_jwk` reverted to its pre-#261 form, the original end-to-end attack reproduces exactly as #256 describes — a real attestation, `cnf.jkt` bound to a key nobody holds:

```
VULNERABLE: the OP minted an attestation for a low-order key nobody holds a private key for:
eyJ0eXAiOiJ3ZWJhbmstYXR0ZXN0K2p3cyIsImFsZyI6IkhTMjU2In0.eyJpc3MiOm51bGwsInN1YiI6InZpY3RpbSIs...
  "cnf": { "jkt": "eV9frzBXPTP92MWWMpoFOh0WI_kJLvGlhcNs15APU_s" }
```

## Changes

1. **Ingest (`parse_public_jwk`)** — a non-Ed25519 `crv` on an OKP key is now a hard refusal instead of a pass-through, and the low-order check runs unconditionally on what remains.
2. **Strict verification** — new `pub(crate) strict_jws` module applies `VerifyingKey::verify_strict` ahead of `jsonwebtoken::decode`.
3. **`client_assertion.rs`** — same gate on `private_key_jwt`. This was in scope item 3 and turned out to be a small delta (the module already routes through `parse_public_jwk` via `select_key`), so it is done rather than deferred.

Error classification follows `authkestra-devsig`: a bad *key* is `BadJwk`, never `ChallengeSignatureInvalid`/`InvalidClientAssertion`. New rejection branches carry `tracing::warn!` per the AGENTS.md DoD.

No new handler was added, so the AGENTS.md axum/actix wiring rule does not apply here.

## Reuse vs. duplication — stated explicitly

`authkestra-devsig` is **not** a dependency of `authkestra-op`, and making it one would invert the layering (devsig is a verifier crate; the OP is the issuer). `authkestra-crypto-util` **is** already a dependency, so the key-parsing half is reused directly via `parse_ed25519_verifying_key_strict`.

The JWS-splitting and `verify_strict` half is a **deliberate port** of `authkestra-devsig/src/eddsa.rs` (#253), documented as such in `strict_jws`'s module doc, with both deviations from the original called out there:

- it *gates* rather than *replaces* `jsonwebtoken::decode`, so claim validation (`exp`/`nbf`/`aud`/alg allow-list) is not duplicated; this is sound because strict verification rejects a strict superset of what non-strict rejects;
- it dispatches on the key being OKP rather than on an `Algorithm` argument, which is exactly the set of keys that reach the EdDSA backend.

Lifting the shared logic into `authkestra-crypto-util` would mean adding `jsonwebtoken` as a dependency of that crate, which is a larger blast radius than the duplication it removes. Flagging it as a possible follow-up rather than doing it here.

## Tests

The reproduction is committed, not just run locally. Note `challenge_signature_verification_itself_refuses_low_order_keys` builds its `Jwk` via `serde_json` rather than `parse_public_jwk`: routing through ingest makes it **vacuous**, because ingest refuses first and the strict gate never executes. I caught this by disabling the gate and finding the test still passed; with the direct construction it fails correctly:

```
VULNERABLE: verify_challenge_signature accepted the universal low-order forgery for crv=Ed25519: Ok("CHAL")
```

Positive controls (a fix that rejects everything is not a fix — the trap flagged in review of #253):

- `genuine_ed25519_key_still_enrols_end_to_end` — full ceremony, real Ed25519 key, attestation minted and `cnf.jkt` verified.
- `challenge_signature_verification_still_accepts_a_genuine_ed25519_signature`
- `still_accepts_an_assertion_under_a_genuine_ed25519_key` — `private_key_jwt` with EdDSA still authenticates.
- `non_okp_keys_are_a_no_op` — EC/RSA verification left entirely untouched.

## Does this warrant `!`? — maintainer decision

I have deliberately **not** decided this. Committed without `!`; happy to amend.

- **Against `!`**: no public API or type changes (`strict_jws` is `pub(crate)`, `OpError` untouched).

> **Correction (raised in review).** The sentence previously here — *"no working configuration breaks … an OKP key with `crv` P-256/P-384/P-521 always died at `compute_cnf_jkt`"* — was **wrong**, and wrong in the direction that matters. It holds only for the *enrolment* path. `verify_client_assertion` never calls `compute_cnf_jkt`, so on `main` a **genuine, non-low-order** Ed25519 key registered under a nonconformant `crv` authenticates successfully today and is refused by this PR:
>
> ```
> main:  CA[genuine-ed25519 crv=P-256] -> Ok("jti-1")
> PR:    CA[genuine-ed25519 crv=P-256] -> Err("... Ed25519 is the only supported OKP curve, got P256")
> ```
>
> That configuration is nonconformant (RFC 8037 mandates `crv: "Ed25519"`) and refusing it is correct — but a working configuration **does** break, so the `!` decision should be made on accurate evidence.
- **For `!`**: #261 — the directly analogous fix, same defect class — shipped as `fix(crypto)!`, and release-plz drives versioning from these messages.

## Verification

All run on the CI-pinned toolchain where relevant; `cargo check --workspace --all-features` included specifically to catch cross-crate breakage that per-crate scoping hides.

| Command | Result |
| --- | --- |
| `cargo fmt --all -- --check` | clean |
| `cargo +1.98.0 clippy -p authkestra-op --all-targets --all-features -- -D warnings` | clean |
| `cargo clippy --workspace --all-features -- -D warnings` | clean |
| `cargo check --workspace --all-features` | clean |
| `cargo test -p authkestra-op --all-features` | 153 passed, 0 failed, 0 ignored |
| `cargo test --workspace --all-features` | all suites pass |

`DOCKER_HOST` was set for the container-backed store tests, so the postgres/mysql/sqlite/redis cases genuinely ran rather than silently skipping.

`Cargo.lock` gains exactly one line: `ed25519-dalek` becomes a direct dependency of `authkestra-op` for `verify_strict`, mirroring #253's rationale and wording in `authkestra-devsig`. It was already in the graph via `jsonwebtoken`'s `rust_crypto` backend and `authkestra-crypto-util`.

## Scope left undone

`crates/authkestra-resource/src/jwt.rs` is untouched, per the issue's own priority ordering and its note that resource is best done after #250/#254. **#256 should not be auto-closed by this PR** if the maintainer considers resource part of it; hence `Refs` rather than `Resolves`.

---

## Corrections and review fixes

An adversarial review found the code sound and could not construct a next bypass, but flagged that this PR **understates what it fixes** and that one of its two gates was untested.

### The client_assertion path is a complete, unmitigated bypass — not "defence in depth"

The description frames `client_assertion.rs` as lower-reachability hardening. Measured on `main`:

```
CA[low-order crv=Ed25519] -> Err("... low-order point ...")   <- #261 catches this
CA[low-order crv=P-256]   -> Ok("jti-1")                      <- full client authentication
```

`verify_client_assertion` never calls `compute_cnf_jkt`, so the accident that saves the enrolment ceremony does not exist on this path. A single low-order key in a client's registered JWKS lets anyone authenticate as that client with a 64-byte constant. This PR closes it; the framing should have said so.

### The `client_assertion` gate was entirely untested — fixed in `213e47c`

Measured: deleting the whole `ensure_strict_eddsa_signature` block from `verify_client_assertion` left the suite at **`153 passed; 0 failed`**. Every existing test routes through `select_key` → `parse_public_jwk`, which refuses these keys at ingest, so the second gate never ran.

This is the same vacuity trap already identified and fixed for the enrolment test — it just wasn't re-applied here. Added `strict_gate_itself_refuses_low_order_keys_for_client_assertions`, built with `serde_json::from_value` so ingest is bypassed on purpose. Verified load-bearing: neutering the gate now fails it, where the same mutation previously left everything green.

### A test's claim corrected — also `213e47c`

`strict_verification_rejects_the_universal_forgery_non_strict_accepts` claimed to pin `verify_strict` over `Verifier::verify`. It doesn't — swapping the call leaves the suite green, because the test asserts a `Key(low-order)` rejection that `is_weak()` emits *before* the signature check. Corrected the comment rather than the code: the only inputs the two verifiers disagree on are a small-order `A` (already caught) and a small-order `R` under a non-small-order `A` (which plain `verify` also rejects), so nothing reachable distinguishes them.

### Also fixes an unclaimed remote panic — evidence for `!`

`from_ed_components` performs no length check, so an `x` shorter than 32 bytes panics at `jsonwebtoken`'s `eddsa.rs:48` (`range end index 32 out of range for slice of length 16`). Reachable through the same `crv`-confusion door on **both** paths, including unauthenticated at the token endpoint. After this PR both refuse at ingest.

### Verification after the fixes

`fmt` clean · `clippy +1.98.0 -p authkestra-op --all-targets` clean · `cargo check --workspace --all-features` clean · **154 passed; 0 failed**.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -631,6 +631,7 @@ dependencies = [
  "authkestra-resource",
  "base64 0.22.1",
  "chrono",
+ "ed25519-dalek",
  "jsonwebtoken",
  "p256",
  "rand 0.9.5",
--- a/crates/authkestra-op/Cargo.toml
+++ b/crates/authkestra-op/Cargo.toml
@@ -21,6 +21,13 @@ async-trait = "0.1"
 chrono = { version = "0.4", features = ["serde"] }
 uuid = { version = "1.0", features = ["v4"] }
 jsonwebtoken = { version = "11", features = ["rust_crypto"] }
+# Direct, not transitive-by-accident: `crate::strict_jws` calls
+# `VerifyingKey::verify_strict` itself because `jsonwebtoken`'s EdDSA backend uses the
+# non-strict verifier (authkestra#256). This adds no new code to the dependency graph —
+# `jsonwebtoken`'s `rust_crypto` backend, already enabled above, resolves the same
+# `ed25519-dalek` 2.x, as does `authkestra-crypto-util` below. Same rationale, and same
+# wording, as `authkestra-devsig`'s equivalent dependency added by #253.
+ed25519-dalek = "2"
 tracing = "0.1.44"
 rand = { workspace = true }
 base64 = "0.22.1"
--- a/crates/authkestra-op/src/attestation.rs
+++ b/crates/authkestra-op/src/attestation.rs
@@ -282,10 +282,41 @@ pub fn parse_public_jwk(raw: &Value) -> Result<Jwk, OpError> {
     match &jwk.algorithm {
         AlgorithmParameters::EllipticCurve(_) | AlgorithmParameters::RSA(_) => Ok(jwk),
         AlgorithmParameters::OctetKeyPair(params) => {
-            if params.curve == jsonwebtoken::jwk::EllipticCurve::Ed25519 {
-                authkestra_crypto_util::parse_ed25519_verifying_key_strict(&params.x)
-                    .map_err(|e| OpError::BadJwk(e.to_string()))?;
+            // `crv` is checked first, and non-Ed25519 is a hard refusal rather
+            // than a pass-through (authkestra#256). `jsonwebtoken`'s
+            // `OctetKeyPairParameters::curve` is the *shared* `EllipticCurve`
+            // enum, so `{"kty":"OKP","crv":"P-256"}` deserialises happily into
+            // this variant — while `expected_algorithm` maps every
+            // `OctetKeyPair` to `Algorithm::EdDSA` and
+            // `DecodingKey::from_jwk` sends every `OctetKeyPair` to
+            // `from_ed_components`, both regardless of `crv`. So `x` is
+            // interpreted as Ed25519 bytes whatever `crv` claims, and a
+            // low-order check written as `if crv == Ed25519 { ... }` is
+            // simply skipped by setting `crv` to anything else. Refusing the
+            // key outright is what makes the check below unavoidable.
+            if params.curve != jsonwebtoken::jwk::EllipticCurve::Ed25519 {
+                tracing::warn!(
+                    curve = ?params.curve,
+                    "rejecting OKP jwk: Ed25519 is the only OKP curve this OP verifies, and every \
+                     OKP key is decoded as Ed25519 downstream — accepting another `crv` would let \
+                     a key skip the low-order check while still reaching the Ed25519 verifier \
+                     (authkestra#256)"
+                );
+                return Err(OpError::BadJwk(format!(
+                    "Ed25519 is the only supported OKP curve, got {:?}",
+                    params.curve
+                )));
             }
+
+            authkestra_crypto_util::parse_ed25519_verifying_key_strict(&params.x).map_err(|e| {
+                tracing::warn!(
+                    error = %e,
+                    "rejecting OKP jwk at ingest; a low-order point here would let the OP mint an \
+                     attestation for a key no private key exists for (authkestra#256)"
+                );
+                OpError::BadJwk(e.to_string())
+            })?;
+
             Ok(jwk)
         }
         AlgorithmParameters::OctetKey(_) => Err(OpError::BadJwk(
@@ -428,6 +459,26 @@ pub(crate) fn verify_challenge_signature(compact_jws: &str, jwk: &Jwk) -> Result
     }
 
     let algorithm = expected_algorithm(jwk)?;
+
+    // Strict EdDSA gate, ahead of `jsonwebtoken`'s non-strict backend
+    // (authkestra#256). A no-op for non-OKP keys. Runs before `decode` so a
+    // low-order key or a low-order `R` can never satisfy proof of possession;
+    // see `strict_jws`'s module doc for why this gates rather than replaces.
+    if let Err(rejection) = crate::strict_jws::ensure_strict_eddsa_signature(compact_jws, jwk) {
+        return Err(match rejection {
+            // A bad KEY is not a bad signature: surfacing this as
+            // `ChallengeSignatureInvalid` would hide that the key is the defect.
+            crate::strict_jws::StrictEdDsaRejection::Key(msg) => {
+                tracing::warn!(error = %msg, "enrolment challenge key refused before verification");
+                OpError::BadJwk(msg)
+            }
+            crate::strict_jws::StrictEdDsaRejection::Signature(msg) => {
+                tracing::warn!(error = %msg, "enrolment challenge signature failed strict EdDSA verification");
+                OpError::ChallengeSignatureInvalid
+            }
+        });
+    }
+
     let decoding_key = DecodingKey::from_jwk(jwk).map_err(|e| OpError::BadJwk(e.to_string()))?;
 
     let mut validation = Validation::new(algorithm);
--- a/crates/authkestra-op/src/client_assertion.rs
+++ b/crates/authkestra-op/src/client_assertion.rs
@@ -370,6 +370,36 @@ pub fn verify_client_assertion(
     })?;
 
     let jwk = select_key(jwks, header.kid.as_deref())?;
+
+    // Strict EdDSA gate, ahead of `jsonwebtoken`'s non-strict backend
+    // (authkestra#256). A no-op for non-OKP keys. Lower reachability than the
+    // enrolment path — this key was registered by an administrator rather than
+    // supplied in the request — but it is the same defect, and "registered by
+    // an administrator" is not "provably not low-order".
+    if let Err(rejection) = crate::strict_jws::ensure_strict_eddsa_signature(assertion, &jwk) {
+        return Err(match rejection {
+            // A bad KEY is not a bad assertion: a low-order registered key is an
+            // operator-facing defect in the client's JWKS, not a client that
+            // signed badly.
+            crate::strict_jws::StrictEdDsaRejection::Key(msg) => {
+                tracing::warn!(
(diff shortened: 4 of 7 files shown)


stephane-segning:
## Adversarial review — cryptographic and input-validation correctness

Reviewed with a deliberately low prior: this is a fix to a **bypass of a previous fix** (#261), so "the gate looks airtight" was not accepted as evidence. The reviewer was asked to find the *next* bypass.

**Verdict: sound. No next bypass found.** But the PR understated what it fixes, and one of its two gates was untested. Both addressed in `213e47c`.

### ✅ All four claims verified end-to-end

**#261's gate is bypassable** — confirmed by reading every link in `jsonwebtoken-11.0.0`: `AlgorithmParameters` is `#[serde(untagged)]` (`jwk.rs:454`), `OctetKeyPairParameters::curve` is the shared 4-variant `EllipticCurve` (`jwk.rs:340-353`), and `decoding.rs:221` routes **every** `OctetKeyPair` to `from_ed_components` with no `crv` inspection. Reproduced on `main`: P-256/384/521 all `Ok("CHAL")`.

**Saved by an accident** — `thumbprint()` returns `Err(InvalidKeyFormat)` for non-Ed25519 OKP as a *serialisation-shape* guard, no security claim in its contract. And per below, it doesn't cover `client_assertion` at all.

**The new gate closes it** — all three P-curves refused at ingest, nothing legitimate broken except the case corrected below.

**`153 passed; 0 failed`** — reproduced exactly (with `DOCKER_HOST` set; without it 5 container tests fail on the known local artefact). `cargo check --workspace --all-features` clean.

### 🔴 The PR understated its own severity

`client_assertion` was framed as lower-reachability hardening. Measured on `main`:

```
CA[low-order crv=Ed25519] -> Err("... low-order point ...")
CA[low-order crv=P-256]   -> Ok("jti-1")     <- full client authentication
```

`verify_client_assertion` never calls `compute_cnf_jkt`, so the accident saving enrolment doesn't exist there. Corrected in the body.

### 🔧 The `client_assertion` gate was untested — fixed in `213e47c`

Deleting the whole gate left **`153 passed; 0 failed`**. Every test routed through `parse_public_jwk`, which refuses at ingest, so the second gate never ran — the same vacuity trap already fixed for enrolment, not re-applied here. New test bypasses ingest deliberately; verified load-bearing (neutering the gate now fails it).

### 🔧 A test's claim corrected — also `213e47c`

`strict_verification_rejects_the_universal_forgery_non_strict_accepts` claimed to pin `verify_strict`; swapping in `Verifier::verify` leaves the suite green, because the assertion checks a `Key(low-order)` rejection emitted by `is_weak()` *before* the signature check. Comment corrected, code kept — no reachable input distinguishes the two verifiers, so `verify_strict` is genuine belt-and-braces.

### 📌 Also fixes an unclaimed remote panic

Short `x` → panic at `jsonwebtoken`'s `eddsa.rs:48`, reachable unauthenticated at the token endpoint via the same door. Now refused at ingest. Evidence for `!`.

### Negative results — hunted, not found

No second door into EdDSA in `authkestra-op` (both `DecodingKey::from_jwk` sites gated; all four `parse_public_jwk` call sites checked). `x` encoding fuzz — short/empty/33-byte/std-alphabet/padded/unicode/whitespace/non-canonical — all reject cleanly, no panics. JWS shape fuzz — 2-segment, 4-segment, empty sig, trailing dot — all clean, and no parser desync (`strict_jws.rs:104` `rsplitn(2, '.')` matches `decoding.rs:361-362`). `EllipticCurve`/`RSA` pass-through is correct. **No stored-vs-presented key swap** — `CompleteChallengeRequest` carries no JWK; the ceremony re-parses from the store.

### Pre-existing, not introduced here

`attestation.rs:340` and `:604` still assert the workspace pins `jsonwebtoken = "10.4.0"` and that `thumbprint()` *panics*. `Cargo.toml:20` is `"11"`, whose `thumbprint` returns `Err`. The `catch_unwind` is harmless; the comments are stale.

<sub>🤖 Adversarial review by an Opus agent. Bypass chain read link-by-link in upstream source and reproduced on `main`; both fixes independently mutation-verified before posting.</sub>

PR #271, opened 2026-08-28
fix(op): add public constructors for non_exhaustive store types

### LINKED ISSUE #268: authkestra-op 0.6.1: non_exhaustive store types have no public constructor, making the store traits unimplementable downstream
## Summary

In `authkestra-op` 0.6.1, `AuthorizationCode`, `DeviceCodeSession`, `RefreshToken` and `TokenRequest` are `#[non_exhaustive]` but expose no public constructor — while the store traits require downstream implementors to **return** them. That makes `AuthorizationCodeStore`, `DeviceCodeStore` and `RefreshTokenStore` unimplementable outside this crate.

Found while bumping `lightbridge-authz` 0.5.5 → 0.6.1 to pick up the `OpStore` authorization_code seam from #249 (which is present and works — thank you).

## The contradiction

`#[non_exhaustive]` blocks struct-literal construction downstream:

| Type | `#[non_exhaustive]` | Public constructor |
|---|---|---|
| `AuthorizationCode` (`code.rs:15`) | yes | no — `impl` has only `is_expired` |
| `DeviceCodeSession` (`device.rs:21`) | yes | no — `impl` has only `is_expired` |
| `RefreshToken` (`refresh.rs:8`) | yes | no `impl` at all |
| `TokenRequest` | yes | no `impl` at all |

But the traits demand implementors produce them:

```rust
// code.rs:70
async fn consume_code(&self, code: &str) -> Result<Option<AuthorizationCode>, OpError>;
// device.rs:55
async fn get_device_code(&self, device_code: &str) -> Result<Option<DeviceCodeSession>, OpError>;
```

A downstream store reads a row from its own database and must return one of these. There is no way to build one. `TokenResponse` and `TokenErrorResponse` did get `::new()` constructors in 0.6.1 — these four appear to have been missed.

## Reproduction

```
cargo add authkestra-op@0.6.1
```

then implement `AuthorizationCodeStore` over any backing store:

```
error[E0639]: cannot create non-exhaustive struct using struct expression
```

Ten such errors across seven files in our tree, all of this shape.

## Impact

0.6.1 cannot be adopted by any crate that implements these stores. For us this blocks consuming #249, which is the release we wanted.

## Possible fixes

1. **Add `::new()` constructors** for all four, matching what `TokenResponse::new` already does. Smallest change, keeps the `#[non_exhaustive]` intent (additive fields stay non-breaking).
2. **Add builders** if the field counts make positional constructors unpleasant — `AuthorizationCode` in particular has many fields.
3. **Drop `#[non_exhaustive]`** from types that downstream implementors are required to construct, keeping it on response types the crate itself always builds.

Option 1 or 3 seems right — the distinction being whether a type is *produced by* the crate (safe to seal) or *produced for* it by implementors (must be constructible).

Happy to send a PR for option 1 if that is the direction you prefer.

---

*Verified against the published 0.6.1 source in `~/.cargo/registry`, not inferred from the changelog. Investigation was AI-assisted; the trait/constructor mismatch was confirmed by reading the actual `impl` blocks and trait signatures.*

marcjazz:
## Summary

`AuthorizationCode`, `DeviceCodeSession`, `RefreshToken` and `TokenRequest` in `authkestra-op` are `#[non_exhaustive]` but expose no public constructor, while `AuthorizationCodeStore`, `DeviceCodeStore` and `RefreshTokenStore` require downstream implementors to *return* them. That makes these traits unimplementable outside this crate — a downstream store reading a row from its own database has no way to build the value it must return.

This adds `::new()` constructors for all four types, matching the pattern already used by `TokenResponse::new()` / `TokenErrorResponse::new()`:

- `AuthorizationCode::new(code, client_id, redirect_uri, scope, identity, expires_at)` — `used: false`, PKCE/nonce fields default `None`.
- `DeviceCodeSession::new(device_code, user_code, client_id, scope, expires_at, status)` — `last_polled_at: None`.
- `RefreshToken::new(token, client_id, identity, scope, expires_at)`.
- `TokenRequest::new(grant_type)` — every other field defaults `None`.

Since every field on these types is already `pub`, `#[non_exhaustive]` only ever blocked the struct-literal syntax; these constructors unblock full reconstruction (optional fields can still be set directly on the returned value).

Fixes #268

## Test plan

- [x] `cargo test -p authkestra-op --lib` — 148 passed (147 existing + 1 new regression test)
- [x] `cargo clippy -p authkestra-op --all-targets -- -D warnings` — clean
- [x] `cargo fmt -p authkestra-op -- --check` — clean
- [x] Added `non_exhaustive_store_types_are_constructible_via_new`, which constructs all four types the way a downstream `*Store` implementation would need to.

### DIFF
--- a/crates/authkestra-op/src/attestation.rs
+++ b/crates/authkestra-op/src/attestation.rs
@@ -177,6 +177,32 @@ pub struct EnrolmentChallenge {
 }
 
 impl EnrolmentChallenge {
+    /// Creates a new enrolment/re-issuance challenge.
+    ///
+    /// `#[non_exhaustive]` blocks struct-literal construction from outside
+    /// this crate, but `EnrolmentChallengeStore` implementations must be
+    /// able to reconstruct a challenge from their own storage — this is the
+    /// seam that makes that possible (authkestra#268).
+    pub fn new(
+        challenge: String,
+        subject: String,
+        principal_id: String,
+        principal_type: PrincipalType,
+        public_jwk: Value,
+        attributes: Value,
+        expires_at: DateTime<Utc>,
+    ) -> Self {
+        Self {
+            challenge,
+            subject,
+            principal_id,
+            principal_type,
+            public_jwk,
+            attributes,
+            expires_at,
+        }
+    }
+
     /// Returns true if this challenge is expired as of `now`.
     pub fn is_expired(&self, now: DateTime<Utc>) -> bool {
         now >= self.expires_at
--- a/crates/authkestra-op/src/code.rs
+++ b/crates/authkestra-op/src/code.rs
@@ -41,6 +41,36 @@ pub struct AuthorizationCode {
 }
 
 impl AuthorizationCode {
+    /// Creates a new, unused authorization code.
+    ///
+    /// `#[non_exhaustive]` blocks struct-literal construction from outside
+    /// this crate, but `AuthorizationCodeStore` implementations must be able
+    /// to reconstruct a code from their own storage — this is the seam that
+    /// makes that possible (authkestra#268). `code_challenge`,
+    /// `code_challenge_method` and `nonce` start `None` and, since every
+    /// field here is `pub`, can be set directly on the returned value.
+    pub fn new(
+        code: String,
+        client_id: String,
+        redirect_uri: String,
+        scope: String,
+        identity: Identity,
+        expires_at: DateTime<Utc>,
+    ) -> Self {
+        Self {
+            code,
+            client_id,
+            redirect_uri,
+            scope,
+            code_challenge: None,
+            code_challenge_method: None,
+            nonce: None,
+            identity,
+            expires_at,
+            used: false,
+        }
+    }
+
     /// Returns true if this code is expired as of `now`.
     pub fn is_expired(&self, now: DateTime<Utc>) -> bool {
         now >= self.expires_at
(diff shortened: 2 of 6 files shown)


marcjazz:
Addressed review feedback in cbb6f47:

1. **`used` fail-open** — `AuthorizationCode::new` now takes `used: bool` as a required parameter instead of hardcoding `false`. This crate's own reconstruction sites fail closed on this exact flag (`sqlx_store.rs`'s `row.try_get("used").unwrap_or(true)`); a constructor that silently defaulted to "unused" was a footgun for downstream stores reconstructing an already-spent code.

2. **Weak in-crate test** — verified: `#[non_exhaustive]` is a crate-boundary restriction, so the original test inside `#[cfg(test)] mod tests` (same compilation unit) didn't actually exercise the bug — it would've passed even against `pub(crate) fn new`. Moved it to `tests/non_exhaustive_store_type_constructors_tests.rs`, a real integration test compiled as an external crate, which is the file that would actually fail to compile pre-fix.

3. **`EnrolmentChallenge`** — confirmed it has the identical defect (`#[non_exhaustive]`, `impl` with only `is_expired()`, and `EnrolmentChallengeStore` requiring implementors to return it). Added `EnrolmentChallenge::new()` alongside the other four.

All claims were independently verified against the code before making any change. 147 lib tests + 6 integration tests (1 existing + 5 new) pass, clippy and fmt clean.

PR #276, opened 2026-08-28
fix(op)!: enforce PKCE unconditionally for the authorization code grant

### LINKED ISSUE #273: op: enforce PKCE unconditionally for the authorization code grant (OAuth 2.1)
## Summary

Split out of #31 after auditing its actual implementation state (see the closing comment on #31). OAuth 2.1 §4.1 requires PKCE on every authorization code exchange. This codebase implements PKCE correctly, but makes it **opt-in per client** rather than mandatory.

## Evidence

- `crates/authkestra-op/src/handlers/authorize.rs:121` — `if client.require_pkce { ... }` only enforces the `code_challenge`/`S256` check when the client's registration has the flag set. A client with `require_pkce: false` can complete the authorization code flow with no PKCE at all.
- `crates/authkestra-op/src/client.rs:220,244,261,282` — every constructor/default for `ClientRegistration` sets `require_pkce: false`.
- `docs/book/ch04-flows-and-protocols.md:7` ("Mandatory PKCE: ... required for all flows") and `docs/book/ch09-security-and-threat-model.md:21` ("PKCE Mandatory: No OAuth flows without Proof Key for Code Exchange") both describe this as already enforced. It isn't, today.

## Proposed fix

- [ ] Require `code_challenge` + `S256` unconditionally in `authorize.rs`, regardless of `client.require_pkce` (or invert the flag into a deliberately-named escape hatch if a migration path for existing non-PKCE clients is needed).
- [ ] Decide on a migration/versioning story for pre-existing clients registered without PKCE — this is a breaking change for them, in the same spirit as the `!` (breaking) commits already used in this repo for similar hardening (e.g. #261).
- [ ] Confirm `docs/book/ch04...` and `ch09...` are accurate once the change lands (they already describe the target state).

## Why this one first

Small and self-contained — no new dependency, no protocol design, a single conditional in one file plus a client-registration migration decision. Compare to the DPoP work (RFC 9449, separately tracked) or the PQC WebAuthn work (#26 follow-up), both of which need new capability built essentially from scratch.

Related: #31 (being closed as superseded by this issue plus a follow-up DPoP issue)

marcjazz:
## Summary

OAuth 2.1 §4.1 requires PKCE on every authorization code exchange. This codebase implements PKCE correctly, but only enforced it when a client's registration set `require_pkce: true` — which every constructor/default sets to `false`. A client not opted in could complete the entire authorization code flow with no PKCE at all.

## Changes

- **`handlers/authorize.rs`** — `handle_authorize` now always requires `code_challenge` + `code_challenge_method=S256`, regardless of `client.require_pkce`. The old three-way "PKCE optional" branch (which produced different error messages depending on partial PKCE params) collapses into the same two checks previously used only for `require_pkce: true` clients.
- **`handlers/token.rs`** — `handle_authorization_code_grant` now always rejects a stored code with no `code_challenge` at redemption too, not just when `client.require_pkce` was set. In normal operation `handle_authorize` never stores a challenge-less code after this change, but this closes the same gap for a legacy pre-existing code, or a downstream `OpStore::store_code` override that bypasses `handle_authorize` entirely.
- **`client.rs`** — `ClientRegistration::require_pkce` is retained on the struct (removing it would be an API-breaking field removal, a separate concern) but is no longer read by either handler; its doc comment now says so.

## Breaking change

Any client previously relying on `require_pkce: false` to skip PKCE will now be rejected at `/authorize` until it sends a `code_challenge`. This is intentional — it's the compliance gap the issue tracks — but it is a behavior break for existing non-PKCE deployments, hence the `!`.

## Test plan

- [x] `cargo test -p authkestra-op --lib` — 149 passed (145 existing/adjusted + 4 new/updated PKCE-focused tests)
- [x] `cargo test --workspace --all-features` — all green, including adapter (axum/actix) and doctest suites
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] New regression tests:
  - `test_pkce_is_mandatory_even_when_client_does_not_require_it` (authorize.rs) — a `require_pkce: false` client sending zero PKCE params is rejected.
  - `test_pkce_is_mandatory_at_redemption_even_for_a_challenge_less_code` (token.rs) — a challenge-less stored code from a `require_pkce: false` client is rejected at `/token`, covering the code path a code that bypassed `handle_authorize` would take.
  - Existing PKCE tests (`test_state_encoding`, `test_pkce_method_without_challenge_redirect_error`, `test_authorization_code_default_behavior_is_unchanged_without_override`) updated with real PKCE fixtures where the scenario they actually cover is orthogonal to PKCE.

Fixes #273

### DIFF
--- a/crates/authkestra-op/src/client.rs
+++ b/crates/authkestra-op/src/client.rs
@@ -107,6 +107,12 @@ pub struct ClientRegistration {
     /// unconditionally, per OAuth 2.1 §4.1 (authkestra#273) — both
     /// `handlers::authorize` and `handlers::token` enforce it regardless of
     /// this value.
+    #[deprecated(
+        since = "0.7.0",
+        note = "PKCE is mandatory for every client on the authorization code grant, \
+                unconditionally, per OAuth 2.1 §4.1 (authkestra#273). This field is no longer \
+                read by handlers::authorize or handlers::token and has no effect."
+    )]
     pub require_pkce: bool,
     /// Downstream audiences (resources) this client is permitted to target
     /// during token exchange.
@@ -197,6 +203,7 @@ where
 }
 
 #[cfg(test)]
+#[allow(deprecated)] // `require_pkce` (authkestra#273) — these fixtures don't exercise it
 mod tests {
 
     use super::*;
--- a/crates/authkestra-op/src/client.rs
+++ b/crates/authkestra-op/src/client.rs
@@ -102,8 +102,11 @@ pub struct ClientRegistration {
     pub grant_types: Vec<GrantType>,
     /// Scopes this client is permitted to request.
     pub scopes: Vec<String>,
-    /// Whether this client must use PKCE (mandatory for public clients,
-    /// recommended for all — see RFC-003 §7 / OAuth 2.1).
+    /// Retained for wire/storage compatibility, but no longer consulted:
+    /// PKCE is mandatory for every client on the authorization code grant,
+    /// unconditionally, per OAuth 2.1 §4.1 (authkestra#273) — both
+    /// `handlers::authorize` and `handlers::token` enforce it regardless of
+    /// this value.
     pub require_pkce: bool,
     /// Downstream audiences (resources) this client is permitted to target
     /// during token exchange.
--- a/crates/authkestra-op/src/client_assertion.rs
+++ b/crates/authkestra-op/src/client_assertion.rs
@@ -470,6 +470,7 @@ pub fn verify_client_assertion(
 }
 
 #[cfg(test)]
+#[allow(deprecated)] // `require_pkce` (authkestra#273) — these fixtures don't exercise it
 pub(crate) mod tests {
     use super::*;
     use crate::client::{GrantType, TokenEndpointAuthMethod};
--- a/crates/authkestra-op/src/handlers/authorize.rs
+++ b/crates/authkestra-op/src/handlers/authorize.rs
@@ -174,6 +174,7 @@ pub async fn handle_authorize(
 }
 
 #[cfg(test)]
+#[allow(deprecated)] // `require_pkce` (authkestra#273) — these fixtures don't exercise it
 mod tests {
     use super::*;
     use crate::client::{ClientRegistration, GrantType};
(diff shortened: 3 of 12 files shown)


marcjazz:
Addressed the review in ebde4ec:

**Discovery metadata (the actionable finding)** — \`OidcDiscovery\` never published \`code_challenge_methods_supported\` at all, so a spec-conformant client had no way to learn from discovery that PKCE is now unconditionally required. Added the field, advertising \`["S256"]\` whenever the provider serves the \`authorization_code\` grant, mirroring the existing conditional-omission pattern already used for \`authorization_endpoint\`. 13 discovery tests pass (11 existing + 2 new), full workspace suite/clippy/fmt clean.

**"Uncommitted changes" note** — checked, and this doesn't apply: the working tree was clean and on the correct branch when I picked this back up. That AtomicInsert/DPoP work is fully committed and pushed as #277, not stray uncommitted state.

**Scope-validation gap** — confirmed real (verified \`handle_authorize\` never checks \`req.scope\` against \`client.scopes\`, including the exact step-numbering fingerprint you pointed at). Filed separately as #278 rather than folding it into this PR, since it's pre-existing and unrelated to PKCE enforcement.

marcjazz:
Correction acknowledged — the two LOW items were still open, not the housekeeping/scope notes. Both addressed in 349cf4c:

**\`error_description\` assertion added** to \`test_pkce_is_mandatory_at_redemption_even_for_a_challenge_less_code\`. Confirmed \`invalid_grant\` alone is produced by 6 other rejection paths in \`default_handle_authorization_code\` — the test now also asserts \`error_description == "PKCE is required"\`, the string unique to the branch it exists to protect.

**\`#[deprecated]\` added** to \`require_pkce\`. This rippled further than a one-liner once the compiler was the judge of it: 12 files reference the field (test fixtures across 6 files in \`authkestra-op\`, the external visibility test, \`sqlx_store.rs\`'s row-mapping, and 4 examples). Each got \`#[allow(deprecated)]\` at the narrowest scope that covers it — module-level for test modules, function-level for the sqlx macro body, statement-level for the example call sites. None needed a behavior change, just acknowledgment that the field is inert now. Verified zero remaining deprecation warnings workspace-wide.

151 \`authkestra-op\` tests + full workspace suite + clippy \`-D warnings\` + fmt all clean.

Thanks for flagging #278's severity — noted for when it's picked up.

Issue #274, opened 2026-08-28
op: implement DPoP (RFC 9449) sender-constrained access tokens


marcjazz:
## Summary

Split out of #31 after auditing its actual implementation state. #31 bundled DPoP together with nonce handling, PKCE enforcement, resource indicators, and deprecated-grant removal — everything except DPoP itself and mandatory PKCE (tracked separately in #273) turned out to already be implemented. This issue tracks the one substantial piece that is genuinely missing: RFC 9449.

## Verified current state

- No `.rs` file in the workspace references DPoP in any form.
- `plans/ticket-31-plan.md` lays out a DPoP task list (proof validation, `TokenResponse` binding, JWK thumbprint binding) but every checkbox is unchecked, and the file hasn't been touched since 2026-05-25.
- `docs/book/ch04-flows-and-protocols.md:9` ("Sender-Constraint: Native support for DPoP...") and `docs/book/ch09-security-and-threat-model.md:23` ("Sender-Constrained Tokens: Using DPoP to bind tokens...") both describe DPoP as already shipped. It is not — these should be corrected once this lands (or sooner, since claiming a security control that doesn't exist is worse than ordinary doc drift).

## Scope (RFC 9449)

- [ ] Validate `DPoP` proof JWTs presented at `/token` (`htm`, `htu`, `iat`, `jti` replay check).
- [ ] Bind issued access tokens to the proof's JWK thumbprint (`cnf.jkt`) — the crate already has a thumbprint-binding pattern for device-signature attestations (`authkestra-op/src/attestation.rs`'s `compute_cnf_jkt`) that this can likely follow/reuse.
- [ ] Reject non-DPoP presentation of a DPoP-bound token at the resource server (`authkestra-resource`).
- [ ] `ath` claim binding (hash of the access token) on resource requests.
- [ ] Optional: server-provided DPoP-nonce challenge (RFC 9449 §8) if replay protection via `iat`+`jti` alone is judged insufficient.

## Out of scope here

- Making PKCE mandatory-by-default — tracked in #273.

Related: #31 (closed, superseded by this issue and #273)

PR #277, opened 2026-08-28
feat(engine): DPoP (RFC 9449) proof verification and atomic replay-guard primitive

marcjazz:
## Summary

Phase A of #274 (DPoP / RFC 9449). This targets the `next` branch, not `main` — per the release-timing discussion on #276, breaking/staged work is accumulating on `next` for now rather than triggering release-plz on every merge.

This lays the foundation both `authkestra-op` (issuance) and `authkestra-resource` (verification) will build on in follow-up PRs. It's landed on its own because it's fully unit-testable in isolation, adds no HTTP-layer wiring, and changes the behavior of nothing that exists today.

## What's here

- **`authkestra-engine::store`** — new `AtomicInsert<T>` trait, the complement of the existing `AtomicConsume<T>`. `AtomicConsume` is for values *this server* creates and later atomically fetches-and-removes (an authorization code). `AtomicInsert` is for values whose key is *client-supplied* and never stored by this server first — exactly what a replay guard needs, since a DPoP proof's `jti` is client-generated. Implemented for all three backends:
  - `MemoryStore` — `HashMap::entry` under the existing mutex.
  - `RedisStore` — `SET key value NX EX ttl`, one round trip.
  - `SqlKvStore<Postgres|Sqlite>` — `INSERT ... ON CONFLICT DO NOTHING`; `SqlKvStore<MySql>` — `INSERT IGNORE`.

- **`authkestra-crypto-util`** — new `verify_ed25519_signature_strict`, a JWS-framing-agnostic sibling of the existing `parse_ed25519_verifying_key_strict`. It exists so more than one call site can share the actual strict-EdDSA check (the #242/#256 universal-forgery defense) instead of each re-deriving it. **`authkestra-op::strict_jws` is untouched** — it's already adversarially reviewed and hardened, and there's no reason to touch it for this.

- **`authkestra-engine::token::dpop`** — `verify_dpop_proof`, a pure function verifying a DPoP proof JWT per RFC 9449 §4.3: `typ`, algorithm derived from the embedded key (never trusted from the header), the strict EdDSA gate above for Ed25519 keys, `htm`/`htu` binding, `iat` freshness, and optional `ath` binding — plus `compute_jwk_thumbprint` for the `cnf.jkt` value. Verified directly against the pinned `jsonwebtoken = "11.0.0"` source that its `thumbprint()` no longer panics on a structurally-inconsistent key (that was specific to the 10.4.0-era implementation `authkestra-op`'s own `compute_cnf_jkt` still guards against), so this doesn't need a `catch_unwind`.

  Deliberately **not** included here (left to callers in Phase B/C): replay checking via the new `AtomicInsert`, binding `cnf.jkt` into an issued token, and comparing a proof's `jkt` against an access token's `cnf.jkt`.

## Follow-up (tracked on #274, not in this PR)

- Phase B: wire into `authkestra-op`'s `/token` endpoint across every grant type (authorization_code, refresh_token, client_credentials, token_exchange), stamping `cnf.jkt` and switching `token_type` to `"DPoP"`.
- Phase C: wire verification into `authkestra-resource`'s `JwtStrategy`.

## Test plan

- [x] `cargo test -p authkestra-engine --lib --features memory,redis,sql-sqlite,sql-postgres,sql-mysql` — 99 passed, including real Redis/Postgres/MySQL integration tests via testcontainers (not just sqlite).
- [x] `cargo test -p authkestra-crypto-util` — 6 passed.
- [x] `cargo test --workspace --all-features` — all green, zero regressions anywhere else.
- [x] `AtomicInsert` concurrency test: 50 tasks racing to insert the same key on `MemoryStore` — exactly one wins, every time.
- [x] `verify_dpop_proof` tests cover: a genuine fresh proof (and correct `jkt`); case-insensitive `htm`; `htu` ignoring query/fragment; stale and future-dated `iat`; wrong `typ`; `ath` match/mismatch/missing-when-expected; a low-order-key universal forgery (mirroring #256's exact attack shape); a smuggled private JWK member; a missing `jwk`; and a known RFC 7638 thumbprint test vector (the literal example from RFC 7638 §3.1).


marcjazz:
All six findings verified against the code first, then fixed. Pushed as 7793bf7.

**HIGH — Redis sub-second TTL bypass.** Confirmed: `ttl.as_secs()` truncates anything under 1s to 0, which hit an early-return reporting `Ok(true)` without ever calling Redis — so a replay of the same key under a sub-second TTL also found nothing stored and *also* reported fresh. Fixed by flooring to Redis EX's own 1-second granularity and always issuing the real `SET NX EX`. Added two regression tests (10ms and exactly `Duration::ZERO`) proving a replay is still blocked.

**MEDIUM — SQL unbounded growth.** Postgres/Sqlite now use `ON CONFLICT DO UPDATE ... WHERE expires_at <= $N` to reclaim an expired row instead of blocking on it forever. MySQL was more interesting: my first pass mirrored the same idea with `ON DUPLICATE KEY UPDATE` + `IF()`, and it looked right by the same reasoning that worked for the other two dialects — but running it against a real MySQL via testcontainers caught it lying: a conflict against a still-*valid* row was still reporting `rows_affected() > 0`. Rewrote MySQL's impl to reuse this same file's existing `SELECT ... FOR UPDATE` transaction pattern (already established for MySQL's `consume`) instead of trusting affected-rows semantics at all. Glad this got caught by an actual database rather than shipped on the strength of "the query reads correctly."

**MEDIUM — DPoP test coverage.** Added `p256`/`rand_core` as `authkestra-engine` dev-dependencies (mirroring `authkestra-op`'s identical existing use) to sign a genuine ES256 proof end-to-end — the algorithm most real DPoP clients actually use, previously completely untested. Also wired up `alg_override` (which existed but nothing called) into a parametrized test covering `none`/`None`/`NONE`/`HS256`/`hs256`/`Hs384`/`HS512`.

**LOW — case sensitivity, jti length, htu doc.** All three fixed: the HS* check now uses `eq_ignore_ascii_case` like `none` already did; `jti` is capped at 255 bytes (matching MySQL's `VARCHAR(255)` key column) and rejected before any store is involved; the `canonicalize_htu` doc no longer claims a lowercasing fallback the code never did.

Re-ran the full workspace suite now that there's disk space (52G free) — clean. 109 engine tests (up from 87), 6 crypto-util, 15 SQL-dialect tests across all three backends via testcontainers, clippy `-D warnings` and fmt both clean workspace-wide.

marcjazz:
Both addressed, pushed as dedf00d.

**HIGH — MySQL deadlock.** Confirmed the mechanism: `SELECT ... FOR UPDATE` on a non-matching row still takes a gap lock under InnoDB's default REPEATABLE READ, so two racers on the same absent key each hold a gap lock and each try to insert into the gap the other holds — classic deadlock. Dropped the transaction entirely. Two separate, individually-autocommitted statements are sufficient and deadlock-free: `INSERT IGNORE` claims a genuinely absent key in one step; only on conflict does a guarded `UPDATE ... WHERE expires_at <= ?` ask whether the existing row is expired, and that statement's own `WHERE` re-evaluates under its own lock — so two callers racing to reclaim the same expired row still can't both win, no explicit transaction required.

Added the concurrency test the deadlock would have failed: 32 racers on one key, asserting no `Err` and exactly one winner. Ran the full MySQL suite 4 times against fresh testcontainers (a 5th was cut off by my own time budget, not a failure) — zero deadlocks across 128 concurrency-test executions.

**Flaky reclaim test.** Confirmed: the migration's `TIMESTAMP` column has no fractional-second precision, so MySQL *rounds* (doesn't truncate) a sub-second value to the nearest whole second — a 1ms TTL's stored `expires_at` can land up to a full second after the actual insert. Widened the sleep to 1100ms rather than touching the column precision, which would affect every consumer of the migration, not just this test. Zero flakes across the same 4 runs.

Thanks for reproducing the deadlock against real MySQL rather than accepting that the query "reads correctly" — that's twice now this PR's SQL logic looked right on paper and wasn't.

marcjazz:
All three LOW items fixed, plus the concurrency test ported to Postgres and Sqlite. Pushed as e2970b8.

**Redis floor-vs-round-up.** Confirmed: `.as_secs().max(1)` only helped the exact-zero case — a TTL like 1.9s still truncated to 1s, silently shortening the replay window on every sub-second-remainder TTL. Now rounds up on any fractional remainder before flooring. Added a test asserting Redis' own `TTL` command reads back 2 for a 1.5s insert, not 1.

**Stale MySQL comment.** Confirmed — it still described the `ON DUPLICATE KEY UPDATE` affected-rows quirk from the implementation the deadlock fix replaced. Corrected to describe the actual two-statement shape.

**Untested `Malformed` branches.** Added coverage for all six production sites: the four segment-count shapes (empty, one segment, two segments, four segments) plus both content-level cases (bad base64url, valid base64url but not JSON).

**Concurrency tests ported to Postgres and Sqlite.** Agreed with the framing — three bugs in this file so far all came from reasoning about what the SQL *says* rather than what the engine does under concurrency, and Postgres's `ON CONFLICT DO UPDATE ... WHERE` was still only asserted by inspection. Both new tests (32 racers, exactly one winner, no errors) pass; ran 3 consecutive times to check for flakiness given MySQL's history in this file — clean every time.

Full gate: 116 engine tests (up from 93), full workspace suite, clippy `-D warnings`, and fmt all clean.

marcjazz:
Applied the trivial nit too — pushed as 133922e, \`saturating_add\` instead of \`+\`. Confirmed the overflow was real (\`Duration::MAX.as_secs() == u64::MAX\`), and the fix needs no dedicated test — it's correct by construction, and meaningfully testing it would mean pushing an absurd EX value at Redis that would likely fail for unrelated reasons anyway. Redis suite (7 tests), clippy, and fmt all still clean.

Appreciate the mutation testing on the concurrency tests — that's exactly the right bar for tests whose entire purpose is catching TOCTOU, and good to have the SQLite doubt (current-thread runtime, in-memory DB) resolved with evidence rather than assumption.

Nothing else outstanding on my end for #277.

PR #279, opened 2026-08-28
fix(op)!: enforce requested scope against the client's registration

### LINKED ISSUE #278: op: /authorize never validates requested scope against the client's registration
## Summary

`handlers::authorize::handle_authorize` never validates the requested `scope` parameter against either the client's registered `scopes` or the server's `scopes_supported`. A client can request — and receive an authorization code, and eventually a token, for — any scope string at all, regardless of what it's registered for.

Surfaced during review of #276 (mandatory PKCE), which doesn't touch this path — this is pre-existing and unrelated to that change.

## Evidence

- `crates/authkestra-op/src/handlers/authorize.rs:147` — `scope: req.scope.clone()` copies the caller-supplied `scope` string verbatim into the stored `AuthorizationCode`, with no check against anything.
- `crates/authkestra-op/src/client.rs:104` — `ClientRegistration::scopes: Vec<String>` exists and is populated on every registration, but nothing in `authorize.rs` ever reads it.
- `ClientRegistration` has `allows_redirect_uri` and `allows_grant_type` (both enforced in `handle_authorize`) but no `allows_scope` equivalent.
- `OpConfig::scopes_supported` (the server-wide advertised scope list, published in discovery) is likewise never consulted at `/authorize`.
- **Textual fingerprint**: `handle_authorize`'s step comments jump from `// 2. Validate exact redirect_uri` straight to `// 4. Check response_type == "code"` — there is no `// 3.` anywhere in the function, suggesting a scope-validation step was either planned and never implemented, or removed at some point without renumbering.

## Impact

A client registered with, say, `scopes: ["profile"]` can request `scope=admin` (or any other scope, including one no client is registered for) at `/authorize` and receive a code — and later a token — carrying that scope. Scope is meant to be the primary access-control boundary between what a client is permitted to do and what it asks for; today that boundary doesn't exist at the authorization code grant's entry point.

## Suggested fix

- Add `ClientRegistration::allows_scope(&self, requested: &str) -> bool` (or an equivalent that checks each space-delimited scope in the request), mirroring `allows_redirect_uri`/`allows_grant_type`.
- In `handle_authorize`, insert a `// 3.` step between the existing `// 2.` and `// 4.` that rejects (via `error_redirect("invalid_scope", ...)`, per RFC 6749 §4.1.2.1) any requested scope not in the client's registered `scopes`.
- Confirm the same gap doesn't exist in the token-exchange / client_credentials / refresh_token scope-handling paths in `handlers/token.rs` (a quick audit, not assumed) — this issue is scoped to `/authorize` specifically since that's where it was found, but the pattern should be checked elsewhere too.

marcjazz:
## Summary

Fixes #278. `handle_authorize` never validated the requested `scope` against either the client's registered `scopes` or the server's `scopes_supported` — it copied `req.scope` verbatim into the stored `AuthorizationCode`, and from there into the eventual token, with no check at all. A client registered for `scopes: ["profile"]` could request `scope=admin` and receive a code, then a token, carrying it.

Unlike #273 (PKCE), this needs **no misconfiguration** to exploit — every deployment has this gap regardless of how clients are registered.

## Audit of other grant paths (per the issue's own ask)

- `default_handle_client_credentials` and `default_handle_token_exchange` **already validate scope correctly** against `client.scopes` — confirmed by reading both, left untouched.
- `default_handle_refresh_token` never reads `req.scope` at all (no narrowing support in this implementation) — it only re-issues whatever scope was already granted, so it has no additional attack surface once the source of that scope is validated.
- `handle_device_authorization` had the **identical gap** to `/authorize` — same root cause (RFC 8628's equivalent of the same request shape), so fixed alongside it rather than filed as a separate issue.

## Changes

- `ClientRegistration::allows_scope(&self, scope: &str) -> bool` — mirrors the existing `allows_redirect_uri`/`allows_grant_type` shape.
- `handle_authorize`: a `// 3.` enforcement step, filling the exact gap the missing step numbering pointed at (steps jumped `2` → `4`). Rejects on the first ungranted scope, naming it specifically via `invalid_scope`, matching `client_credentials`'s existing error shape.
- `handle_device_authorization`: the equivalent check.

## Breaking change

Like #273, this is intentional but breaking: any client currently receiving scopes outside its registration will be rejected until its registration is corrected or the request is narrowed. Targeting `next`, not `main`, for the same release-timing reason as #273/#277.

**Design choice, explicitly**: RFC 6749 §3.3 permits an AS to *narrow* an over-broad scope request to the granted subset instead of rejecting it outright, and some providers do exactly that. This PR rejects the whole request instead — matching what `client_credentials` already does here, for consistency across grant types — which is the more disruptive option for any existing deployment currently relying on silent narrowing (there was none before this PR, but a future scope-narrowing change would be a *smaller* behavior shift than what this PR ships).

## Test plan

- [x] `cargo test -p authkestra-op --lib` — 154 passed (151 existing + 3 new).
- [x] `cargo test --workspace --all-features` — full suite green.
- [x] Fixed 7 existing `authorize.rs` tests whose fixtures registered a client with `scopes: []` while requesting `"openid"`/`"openid profile"` — updated to request scopes they're actually registered for, so each test again exercises the concern it was written for (PKCE, response_type, state encoding) rather than being blocked by the new scope check.
- [x] New regression tests:
  - `test_scope_not_granted_to_client_is_rejected` (authorize.rs) — a `scopes: ["profile"]` client requesting `scope=admin` is rejected.
  - `test_one_ungranted_scope_among_several_is_rejected` (authorize.rs) — one bad scope among several granted ones still fails the whole request.
  - `test_scope_not_granted_to_client_is_rejected` (device_authorization.rs) — the same scenario for the device flow.

### DIFF
--- a/crates/authkestra-op/src/client.rs
+++ b/crates/authkestra-op/src/client.rs
@@ -154,6 +154,16 @@ impl ClientRegistration {
         self.grant_types.contains(grant_type)
     }
 
+    /// Checks if the client is registered for a single scope value.
+    ///
+    /// Takes one already-split scope token, not a space-delimited string —
+    /// callers validating a whole `scope` request should split it
+    /// themselves and check each token, so they can name the specific
+    /// offending scope in their own error response (authkestra#278).
+    pub fn allows_scope(&self, scope: &str) -> bool {
+        self.scopes.iter().any(|s| s == scope)
+    }
+
     /// Verifies the provided secret against the stored argon2 hash in constant time.
     pub fn verify_secret(&self, secret: &str) -> bool {
         if let Some(hash_str) = &self.client_secret_hash {
--- a/crates/authkestra-op/src/handlers/authorize.rs
+++ b/crates/authkestra-op/src/handlers/authorize.rs
@@ -92,6 +92,23 @@ pub async fn handle_authorize(
         AuthorizeOutcome::Redirect(url.into())
     };
 
+    // 3. Validate requested scope against the client's registration
+    // (authkestra#278). `req.scope` was previously copied verbatim into the
+    // stored code — and from there into the eventual token — with no check
+    // at all, letting any registered client request (and receive) a scope
+    // it was never granted. Rejects on the first offending scope, naming it
+    // specifically, mirroring `default_handle_client_credentials`'s
+    // existing (and correct) scope check at the token endpoint.
+    for s in req.scope.split_whitespace() {
+        if !client.allows_scope(s) {
+            tracing::warn!(client_id = %req.client_id, scope = %s, "Client requested unauthorized scope");
+            return error_redirect(
+                "invalid_scope",
+                &format!("Scope {s} is not allowed for this client"),
+            );
+        }
+    }
+
     // 4. Check response_type == "code"
     if req.response_type != "code" {
         tracing::warn!(
@@ -292,7 +309,7 @@ mod tests {
                     client_secret_hash: None,
                     redirect_uris: vec!["https://app.example.com/cb".to_string()],
                     grant_types: vec![GrantType::AuthorizationCode],
-                    scopes: vec![],
+                    scopes: vec!["openid".to_string()],
                     require_pkce: false,
                     allowed_audiences: vec![],
                     token_endpoint_auth_method: None,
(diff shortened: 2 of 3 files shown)


marcjazz:
Both findings addressed:

**Design-choice wording** — updated the "Breaking change" section to state explicitly that rejecting the whole request (rather than narrowing to the granted subset, which RFC 6749 §3.3 also permits) is the more disruptive option, and why this PR takes it anyway (consistency with `client_credentials`'s existing behavior).

**Missing-scope 400** — verified: `AuthorizeRequest::scope` is a mandatory `String`, and both `Query<AuthorizeRequest>` extractors (axum and actix) fail deserialization on a missing key with a generic 400, not an OAuth-shaped error — even though RFC 6749 §3.3 makes `scope` optional. Filed as #280 rather than fixed here, since it's pre-existing and unrelated to this PR's scope-enforcement change.

Thanks for the thorough audit table on the token-issuance sites — good confirmation the fix is complete.

PR #281, opened 2026-08-28
fix(op): make AuthorizeRequest::scope default to empty per RFC 6749 §3.3

### LINKED ISSUE #280: op: /authorize's scope field is mandatory, but RFC 6749 §3.3 makes it optional
## Summary

`AuthorizeRequest::scope` is typed `String`, not `Option<String>`, but RFC 6749 §3.3 says: *"If the client omits the scope parameter when requesting authorization, the authorization server MUST either process the request using a pre-defined default value, or fail the request indicating an invalid scope."* Either way, a **missing** `scope` is meant to be a normal, spec-legal request shape that `handle_authorize` gets to decide how to handle — not something that fails before `handle_authorize` is even called.

Surfaced during review of #279 (mandatory scope-registration enforcement), which doesn't touch this — pre-existing and unrelated to that change.

## Evidence

- `crates/authkestra-op/src/handlers/authorize.rs:22` — `pub scope: String,` (no `Option`, no `#[serde(default)]`).
- `crates/authkestra-axum/src/op.rs:67` — `Query(req): Query<authkestra_op::handlers::authorize::AuthorizeRequest>` — axum's `Query` extractor deserializes the query string via serde; a mandatory non-`Option` field with no default fails deserialization if the key is absent.
- `crates/authkestra-actix/src/op.rs:41` — `req: web::Query<AuthorizeRequest>` — same extractor pattern, same failure mode, on actix.

## Impact

A client that omits `scope` entirely (spec-legal — e.g. relying on a server-side default) never reaches `handle_authorize` at all. Both `Query` extractors reject the deserialization failure with a generic framework 400 Bad Request, not a proper OAuth2 error response — no `error=invalid_request` redirect (impossible anyway, since `client_id`/`redirect_uri` haven't been validated yet at that point) and no OAuth-shaped body. This is a conformance gap: RFC 6749 clients that correctly omit an optional parameter get an ungraceful failure instead of a spec-compliant one.

## Suggested fix

- Change `AuthorizeRequest::scope` to `Option<String>` (with `#[serde(default)]` so a missing key deserializes to `None` rather than erroring).
- Decide `handle_authorize`'s behavior for `None`: either treat it as an empty-scope request (matches "pre-defined default value" of "no scope"), or apply some other configured default. RFC 6749 leaves this to the AS's own policy.
- Update every call site currently treating `req.scope` as a plain `String` (`handle_authorize` itself post-#279, tests, and any consumer code) accordingly.

marcjazz:
## Summary

Fixes #280. RFC 6749 §3.3: *"If the client omits the scope parameter when requesting authorization, the authorization server MUST either process the request using a pre-defined default value, or fail the request indicating an invalid scope."* Either way, a missing `scope` is a normal, spec-legal request shape — not something that should fail before `handle_authorize` even runs.

`AuthorizeRequest::scope` was typed `String` with no `#[serde(default)]`. Both `axum`'s and `actix`'s `Query<AuthorizeRequest>` extractors deserialize the query string through the same derived `Deserialize` impl, and a mandatory field with no key present and no default fails deserialization outright — before `client_id`/`redirect_uri` are even validated, so there's no way to redirect a proper OAuth error at that point. The client got a raw framework 400 instead of a conformant response.

## Changes

- `AuthorizeRequest::scope` gets `#[serde(default)]`, staying a plain `String`. A missing key now deserializes to `""` instead of failing.

**Revision note**: this PR originally changed the field to `Option<String>` instead. Review caught that the `Option` was never actually used to distinguish "omitted" from "explicitly empty" anywhere — both collapsed to the same `unwrap_or_default()` treatment at every call site — which is exactly the signal that `#[serde(default)]` on the plain `String` was the right shape: identical fix for the actual bug, no breaking type change, no fixture churn. Reverted to that shape.

## No longer a breaking change

Unlike #273/#279, this one isn't breaking: the field's type is unchanged, so nothing constructing `AuthorizeRequest` needs to change.

## Test plan

- [x] `cargo test -p authkestra-op --lib` — 156 passed.
- [x] `cargo test --workspace --all-features` — full suite green.
- [x] `deserializes_successfully_when_scope_is_entirely_absent` — deserializes an `AuthorizeRequest` from a JSON object with no `scope` key at all (a dependency-free stand-in for a query string missing the parameter, exercising the same derived `Deserialize` impl the `Query` extractors use) and asserts `scope == ""`, rather than failing as it did before.
- [x] `test_missing_scope_is_treated_as_no_scope_and_succeeds` — the end-to-end half: an `AuthorizeRequest` built via the same deserialization path (not a struct literal) with no `scope` key, run through `handle_authorize`, succeeds with a persisted code carrying `scope == ""`.

### DIFF
--- a/crates/authkestra-op/src/handlers/authorize.rs
+++ b/crates/authkestra-op/src/handlers/authorize.rs
@@ -18,8 +18,12 @@ pub struct AuthorizeRequest {
     pub redirect_uri: String,
     /// Response type (must be "code").
     pub response_type: String,
-    /// Space-delimited scopes requested.
-    pub scope: String,
+    /// Space-delimited scopes requested. `None` when the client omits the
+    /// parameter entirely — RFC 6749 §3.3 makes this legal, and
+    /// `handle_authorize` treats it the same as an explicitly empty scope
+    /// (authkestra#280): no scope requested, not an error and not
+    /// distinguished from `Some(String::new())`.
+    pub scope: Option<String>,
     /// Optional opaque state parameter.
     pub state: Option<String>,
     /// PKCE code challenge.
@@ -99,7 +103,7 @@ pub async fn handle_authorize(
     // it was never granted. Rejects on the first offending scope, naming it
     // specifically, mirroring `default_handle_client_credentials`'s
     // existing (and correct) scope check at the token endpoint.
-    for s in req.scope.split_whitespace() {
+    for s in req.scope.as_deref().unwrap_or_default().split_whitespace() {
         if !client.allows_scope(s) {
             tracing::warn!(client_id = %req.client_id, scope = %s, "Client requested unauthorized scope");
             return error_redirect(
@@ -161,7 +165,7 @@ pub async fn handle_authorize(
         code: code_val.clone(),
         client_id: client.client_id.clone(),
         redirect_uri: req.redirect_uri.clone(),
-        scope: req.scope.clone(),
+        scope: req.scope.clone().unwrap_or_default(),
         code_challenge: req.code_challenge.clone(),
         code_challenge_method: req.code_challenge_method.clone(),
         nonce: req.nonce.clone(),
@@ -235,7 +239,7 @@ mod tests {
             client_id: "unknown".to_string(),
             redirect_uri: "https://app.example.com/cb".to_string(),
             response_type: "code".to_string(),
-            scope: "openid".to_string(),
+            scope: Some("openid".to_string()),
             state: None,
             code_challenge: None,
             code_challenge_method: None,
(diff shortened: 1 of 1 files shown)


marcjazz:
Good catch — pushed as dd5c2ab, and I don't disagree with the conclusion: the \`Option\` was decoration. I never used it to distinguish "omitted" from "explicitly empty" anywhere in the code — both collapsed to the same \`unwrap_or_default()\` at every call site, which is precisely the argument for \`#[serde(default)] pub scope: String\` instead. Identical fix for the actual bug, no breaking type change, none of the 11 fixture updates.

Reverted the \`Option<String>\` plumbing back to plain \`String\`, retitled/re-described the PR to drop the \`!\`, and updated both regression tests to assert \`scope == ""\`. The deserialization test now doubles as direct proof \`#[serde(default)]\` does its job.

156 tests, full workspace suite, clippy, and fmt all reverified clean.

Please send the agent at #277 whenever you're ready — nothing pending on my end besides that.

PR #286, opened 2026-08-30
feat(op): wire DPoP (RFC 9449) proof verification into the /token endpoint

marcjazz:
## Summary

Phase B of #274. Phase A (#277, merged to `next`) built `authkestra_engine::token::dpop::verify_dpop_proof` (pure proof verification) and `authkestra_engine::store::AtomicInsert` (atomic insert-if-absent), but wired neither into anything. This connects them to `authkestra-op`'s `/token` endpoint: a client presenting a `DPoP` header gets its proof verified, its `jti` checked for replay, and — on success — its issued token bound to the proof's key via `cnf.jkt` with `token_type` switched to `"DPoP"`, across all five grant handlers (`authorization_code`, `refresh_token`, `client_credentials`, `token_exchange`, `device_code`).

Targets `next`, not `main`, per the same release-timing reasoning as #277/#279/#281.

## What's here

- **New `authkestra_op::dpop` module**: `DpopReplayStore` trait, a blanket impl over any backend with `KvStore + AtomicInsert` (Memory/Redis/Postgres/Sqlite/MySQL all get it for free — the first real consumer of Phase A's `AtomicInsert`), and `NoDpopReplayStore` as the fail-closed default (mirrors `NoClientAssertionStore`): a deployment with no replay store wired refuses every DPoP proof rather than accepting one it can't guarantee is single-use.
- **`CompositeOpStore`** gains a 6th defaulted generic slot (`P = NoDpopReplayStore`) and a `with_dpop_replay_store` builder, exactly mirroring the existing `with_client_assertion_store` pattern. `CompositeOpStore::new`'s 4-arg signature is untouched — zero breakage to any existing call site.
- The verified proof's `jkt` is threaded through a new `pub(crate)`, `#[serde(skip)]` field on `TokenRequest` rather than a new trait-method or free-function parameter — both `OpStore` method signatures and the public `default_handle_*` free functions are a pinned, tested contract (`default_handle_token_exchange_visibility_tests.rs`). `#[serde(skip)]` is load-bearing: without it a client could submit `dpop_jkt` as an ordinary form field and bind a token to a key it never proved possession of.
- `handle_token_with_client_cert` gains a trailing `dpop_header` parameter (same precedent as `client_cert_der`), verifies the proof against `config.token_endpoint()` before any grant dispatch, and rejects with `invalid_dpop_proof` on a bad proof, a replay, or no replay store configured.
- A `merge_dpop_cnf` helper folds `cnf.jkt` into any existing `extra` claims — RFC 9449 §6.1 requires `cnf.jkt` and RFC 8705's `cnf.x5t#S256` to coexist in one object, not clobber each other — used at all five issuance sites.
- axum/actix adapters read the `DPoP` header the same way they already read `Authorization`; no new extension-based plumbing needed.
- Discovery gains an opt-in `dpop_signing_alg_values_supported` via `with_dpop_support()`, following the same opt-in precedent as `with_private_key_jwt()`: since the replay store defaults to fail-closed, advertising DPoP unconditionally would be misleading for a deployment that hasn't wired one.

## Deliberate scope boundary

RFC 9449 §4.3's refresh-token key continuity (binding a *rotated* refresh token to the same key the original was bound to, not just any valid proof) is out of scope here — it needs a schema change to `RefreshToken` plus new validation logic, not just wiring. Phase B binds whatever proof is presented on a refresh request to the newly issued token, without enforcing it matches the original. Flagging explicitly rather than silently shipping partial; worth its own follow-up issue.

## Test plan

- [x] `cargo test -p authkestra-op --lib` — 168 passed (up from 160 before this PR, zero regressions).
- [x] New `dpop_wiring_tests` module: valid proof stamps `cnf.jkt` + `token_type: "DPoP"` (client_credentials and authorization_code); no header stays plain `Bearer` with no `cnf` (regression proof the feature is additive); replayed `jti` rejected; `NoDpopReplayStore` wired fails closed; wrong `htu` rejected.
- [x] `cargo test --workspace --all-features` — clean.
- [x] `cargo test -p authkestra-op --test default_handle_token_exchange_visibility_tests` — still passes unchanged, confirming the free-function contract this design relies on is intact.

## Follow-up (tracked on #274, not in this PR)

- RFC 9449 §4.3 refresh-token key continuity (see scope boundary above).
- Phase C: wire verification into `authkestra-resource`'s `JwtStrategy`.

### DIFF
--- a/crates/authkestra-actix/src/op.rs
+++ b/crates/authkestra-actix/src/op.rs
@@ -15,7 +15,7 @@ use authkestra_op::{
             CompleteChallengeRequest, EnrolStartRequest, ReissueStartRequest,
         },
         jwks::JwksResponse,
-        token::{handle_token_with_client_cert, TokenRequest},
+        token::{handle_token_with_client_cert, TokenErrorResponse, TokenRequest},
         userinfo::{handle_userinfo, UserInfoErrorResponse, UserInfoRequest},
     },
     OpError,
@@ -106,6 +106,40 @@ pub async fn actix_device_authorization_handler(
     }
 }
 
+/// Reads the request's `DPoP` header (RFC 9449 §4.1: exactly one is
+/// expected). Returns `Ok(None)` if it's absent, `Ok(Some(value))` if
+/// exactly one occurrence is present and valid ASCII, or `Err` in two
+/// cases that are refused outright rather than tolerated:
+///
+/// - **More than one occurrence.** Which one would be authoritative is
+///   ambiguous, and a duplicate is exactly the kind of situation a
+///   proxy bug or request-smuggling attempt could produce — a
+///   sender-constraining mechanism shouldn't guess which header to
+///   trust.
+/// - **A value that isn't valid ASCII.** Treating this as "no header"
+///   would silently issue a plain Bearer token while the client
+///   believes (because it sent one) that it's getting a
+///   sender-constrained one — a mangled header should fail loudly, not
+///   downgrade silently.
+fn extract_dpop_header(
+    headers: &actix_web::http::header::HeaderMap,
+) -> Result<Option<&str>, TokenErrorResponse> {
+    let mut dpop_headers = headers.get_all("DPoP");
+    match (dpop_headers.next(), dpop_headers.next()) {
+        (None, _) => Ok(None),
+        (Some(_), Some(_)) => Err(TokenErrorResponse::new(
+            "invalid_dpop_proof".to_string(),
+            "Multiple DPoP headers were presented".to_string(),
+        )),
+        (Some(v), None) => v.to_str().map(Some).map_err(|_| {
+            TokenErrorResponse::new(
+                "invalid_dpop_proof".to_string(),
+                "DPoP header value is not valid ASCII".to_string(),
+            )
+        }),
+    }
+}
+
 /// Handler for the token endpoint.
 ///
 /// Certificate binding (RFC 8705, issue #224): this crate does not
@@ -135,8 +169,13 @@ pub async fn actix_token_handler(
 
     // RFC 9449 — an ordinary header, unlike the mTLS certificate above:
     // `http_req` already gives access to the whole header map, no
-    // extension-based plumbing needed.
-    let dpop_header = http_req.headers().get("DPoP").and_then(|h| h.to_str().ok());
+    // extension-based plumbing needed. See `extract_dpop_header` for why
+    // more than one occurrence, or one that isn't valid ASCII, is refused
+    // outright rather than tolerated.
+    let dpop_header = match extract_dpop_header(http_req.headers()) {
+        Ok(h) => h,
+        Err(err) => return HttpResponse::BadRequest().json(err),
+    };
 
     match handle_token_with_client_cert(
         req.into_inner(),
@@ -392,5 +431,54 @@ impl OpActixExt for &mut web::ServiceConfig {
             self.app_data(web::Data::new(provider));
         }
         self
+    }
+}
+
+#[cfg(test)]
+mod dpop_header_tests {
+    use super::extract_dpop_header;
+    use actix_web::http::header::{HeaderMap, HeaderValue};
+
+    #[test]
+    fn no_header_is_none() {
+        let headers = HeaderMap::new();
+        assert_eq!(extract_dpop_header(&headers).unwrap(), None);
+    }
+
+    #[test]
+    fn exactly_one_header_is_returned() {
+        let mut headers = HeaderMap::new();
+        headers.insert(
+            actix_web::http::header::HeaderName::from_static("dpop"),
+            HeaderValue::from_static("proof-value"),
+        );
+        assert_eq!(extract_dpop_header(&headers).unwrap(), Some("proof-value"));
+    }
+
+    /// RFC 9449 §4.1 expects exactly one `DPoP` header; a duplicate is
+    /// refused outright rather than silently resolved by picking one.
+    #[test]
+    fn two_headers_are_rejected() {
+        let mut headers = HeaderMap::new();
+        let name = actix_web::http::header::HeaderName::from_static("dpop");
+        headers.append(name.clone(), HeaderValue::from_static("first"));
+        headers.append(name, HeaderValue::from_static("second"));
+        let err = extract_dpop_header(&headers).expect_err("a duplicate header must be refused");
+        assert_eq!(err.error, "invalid_dpop_proof");
+    }
+
+    /// A header value that isn't valid ASCII must be refused, not silently
+    /// treated as "no header present" (which would issue a plain Bearer
+    /// token while the client believes it asked for — and thinks it got —
+    /// a sender-constrained one).
+    #[test]
+    fn non_ascii_header_is_rejected() {
+        let mut headers = HeaderMap::new();
+        headers.insert(
+            actix_web::http::header::HeaderName::from_static("dpop"),
+            HeaderValue::from_bytes(&[0xff, 0xfe]).unwrap(),
+        );
+        let err = extract_dpop_header(&headers).expect_err("a non-ASCII header must be refused");
+        assert_eq!(err.error, "invalid_dpop_proof");
     }
 }
--- a/crates/authkestra-actix/src/op.rs
+++ b/crates/authkestra-actix/src/op.rs
@@ -133,6 +133,11 @@ pub async fn actix_token_handler(
         .get::<ClientCertificateDer>()
         .map(|c| c.0.clone());
 
+    // RFC 9449 — an ordinary header, unlike the mTLS certificate above:
+    // `http_req` already gives access to the whole header map, no
+    // extension-based plumbing needed.
+    let dpop_header = http_req.headers().get("DPoP").and_then(|h| h.to_str().ok());
+
     match handle_token_with_client_cert(
         req.into_inner(),
         auth_header,
@@ -140,6 +145,7 @@ pub async fn actix_token_handler(
         op_store.get_ref().as_ref(),
         tokens.get_ref().as_ref(),
         client_cert_der.as_deref(),
+        dpop_header,
     )
     .await
     {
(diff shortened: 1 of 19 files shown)


marcjazz:
Thanks — verified all three against the code and pushed fixes in 52569a0.

**HIGH (refresh path), confirmed and fixed.** `RefreshToken` gains an optional `jkt`, set at every mint site (authorization_code, device_code, rotation) from the request's verified proof key. `default_handle_refresh_token` now enforces RFC 9449 §5 continuity: a bound refresh token requires a fresh proof for the *same* key on redemption — mismatch or missing proof → `invalid_dpop_proof`, never a silent fallback to Bearer or re-binding to a new key. The resolved (stored, not client-supplied) key carries forward onto the rotated refresh token and the new access token's `cnf.jkt`. Unbound (pre-DPoP) refresh tokens are unaffected — confirmed via a dedicated regression test. Also updated `sqlx_store.rs`: nullable `jkt` column added to all three SQL backends' schema/insert/read paths, since the SQL-backed store needed the same capability as the in-memory one.

**jti TTL/skew gap, confirmed and fixed.** You're right that `expires_at = Utc::now() + max_age` doesn't track `verify_dpop_proof`'s actual freshness anchor (`iat`). Worst case (an `iat` right at the clock-skew allowance's edge, processed instantly) the record could expire a few seconds before the proof would independently be rejected as stale. Now anchored to `iat + max_age` (+1s boundary buffer), independent of processing delay.

**MemoryStore leak, confirmed and fixed.** Right that `insert_if_absent` never reclaimed a key nothing else revisits — DPoP's replay guard is the first insert-only consumer of `AtomicInsert`, so every proof (e.g. from a polling device-code client) leaked a permanent entry. `insert_if_absent` now sweeps expired entries on every call, bounding the map by roughly the insert rate over one TTL window rather than total inserts ever made. Added a test-only `len()` introspection helper plus a regression test proving the map actually shrinks, not just that expired entries are unreachable via `get`.

One thing I did *not* add: integration-test coverage for the new `jkt` column against real Postgres/MySQL/SQLite — there wasn't any existing testcontainers coverage for `RefreshTokenStore`'s SQL persistence at all (the existing sqlx_store tests only cover `AuthorizationCode` cascading-delete/concurrency), so this is a pre-existing gap, not a regression. The column follows the exact nullable-column idiom already used and covered for `code_challenge`/`code_challenge_method`. Happy to add that suite as a follow-up if you'd rather not carry the gap forward.

Also agreed on the out-of-crate `handle_token_with_client_cert` signature widening being breaking — that's expected and accepted here, same as everything else staged on `next`.

Test plan for this round:
- New tests: same-key rotation succeeds and preserves binding on both tokens; different key refused; missing proof on a bound token refused; unbound refresh token's rotation unaffected; MemoryStore leak-reclaim test.
- `cargo test -p authkestra-op --lib`: 172 passed (up from 168).
- `cargo test -p authkestra-engine --lib --features memory,redis,sql-sqlite,sql-postgres,sql-mysql`: 17 insert_if_absent tests passed, including real Redis/Postgres/MySQL via testcontainers.
- `cargo test --workspace --all-features`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check`: all clean.

marcjazz:
All three confirmed and fixed in 6a2c575 — both of my own last round's fixes had real problems, thanks for catching them.

**HIGH (migration gap), confirmed and fixed.** You're right, and it's exactly the mistake the codebase's own comment on `token_endpoint_auth_method`/`jwks` already warns against — I should have followed that precedent the first time instead of introducing a new instance of the same hazard. Reverted the `jkt` column from every schema, INSERT, and SELECT in `sqlx_store.rs`. A `SqlxOpStore`-backed refresh token now reconstructs with `jkt: None`, with a comment pointing at the same established limitation. `RefreshToken::new`'s `jkt` parameter stays — Memory/Redis-backed deployments still get full continuity — but the SQL backend doesn't persist it until a real ALTER-based migration path exists, same posture as those other two fields.

**Consume-before-validate, confirmed and fixed.** Correct — `consume_token` ran before the DPoP check, so a failed check on an already-deleted token was exactly the inverted outcome you described. Reordered `default_handle_refresh_token` to `get_token` (non-destructive) → validate client_id/expiry/DPoP continuity → `consume_token` only once everything passes. `consume_token`'s atomicity still fully covers the actual replay-prevention property (two concurrent requests can't both succeed), it just no longer runs before we know the request has any right to the token. I mutation-tested the fix: temporarily reintroduced the old ordering, confirmed the new regression test (`refresh_token_survives_a_failed_dpop_attempt_and_a_correct_retry_still_succeeds`) fails against it exactly as expected (`invalid_grant` instead of `invalid_dpop_proof`), then reverted and confirmed it passes again.

**memory.rs:136 quadratic sweep, confirmed and fixed.** Replaced the full-map `retain` with a min-heap of `(expiry, key)` pairs scoped to `insert_if_absent`'s own key space. Each call now pops only entries that have actually expired — amortized O(log n) per call, since each entry is pushed once and popped at most once — and live entries are never touched by the sweep at all, regardless of how many exist. Added a test proving a stale heap entry for a key that gets reused with a fresh TTL is correctly recognized as stale and doesn't wrongly evict the key's current value.

Thanks also for confirming the TTL anchoring, `#[serde(skip)]`/`pub(crate)` on `dpop_jkt`, `expected_algorithm`, `merge_dpop_cnf`, and the `Option` serde default reasoning — no changes needed there.

Test plan:
- New: refresh-token-survives-failed-attempt test (mutation-tested against the old ordering as described above), MemoryStore stale-queue-entry test.
- `cargo test -p authkestra-op --lib`: 173 passed (up from 172).
- `cargo test -p authkestra-engine --lib --features memory,redis,sql-sqlite,sql-postgres,sql-mysql`: 19 `insert_if_absent` tests passed, including real Redis/Postgres/MySQL via testcontainers.
- `cargo build -p authkestra-op --features sqlx-postgres,sqlx-sqlite,sqlx-mysql --tests`: clean.
- `cargo test --workspace --all-features`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check`: all clean.

marcjazz:
All five addressed in d7dd673, with one explicit deferral flagged below.

**1. Migration fix went the wrong way (HIGH) — confirmed.** You're right that reverting persistence just moved the problem: SQL-backed refresh continuity was silently inert. I did *not* build the real ALTER-based migration in this PR — deferring that (Postgres has `ADD COLUMN IF NOT EXISTS`; SQLite/MySQL need an existence check first, which is also exactly what would unblock `token_endpoint_auth_method`/`jwks`, stuck on the same gap) since it's meaningfully more work and testcontainers coverage than fits cleanly here. What I did land: a fail-closed guard. `default_handle_authorization_code`, `handle_device_code`, and `default_handle_refresh_token` now read back any refresh token that's supposed to be DPoP-bound and refuse with `server_error` if the binding didn't survive storage. This closes the *security* gap completely — no code path can any longer issue a token whose response claims a binding storage doesn't have — leaving only an availability gap (SQL-backed deployments can't combine DPoP with `offline_access` yet, loudly, instead of silently). Mutation-tested with a wrapper store that reproduces the exact silent-drop bug. Happy to build the real migration now if you'd rather not carry the gap forward as a follow-up.

**2. SQL TTL truncation (three rounds) — confirmed and fixed.** Extracted the Redis fix into a shared `ttl_ceil_secs` in `authkestra_engine::store`, had Redis and all three SQL backends call it, and rewrote the two tests that depended on the truncation instead of catching it. Also caught a knock-on effect: the MySQL reclaim test's 1100ms sleep margin was tuned to the *old, buggy* short TTL — once the TTL correctly rounds up to a full second minimum, MySQL's own `TIMESTAMP` rounding on top of that pushed the real worst case to ~2s, and the test started failing intermittently in the full workspace run. Widened its margin to 2100ms and confirmed stable across repeated runs.

**3. Adapter header handling (repeat) — confirmed and fixed.** Extracted a testable `extract_dpop_header` in both adapters: a duplicate `DPoP` header or a non-ASCII value is now refused outright with `invalid_dpop_proof`, not silently resolved by picking one or downgrading to Bearer.

**4. `dpop_jkt` visibility (repeat) — confirmed and fixed.** Field stays `pub(crate)` — that's still what makes `#[serde(skip)]` forgery-proof — but added a read-only `TokenRequest::dpop_jkt()` accessor and made `merge_dpop_cnf` `pub`, so an external `OpStore` override can tell whether a request was DPoP-bound and merge its own claims without risking reintroducing the `cnf.x5t#S256` clobber that function exists to prevent.

**5. Resource-server gap — doc warning added**, as scoped. `with_dpop_support()` now explicitly documents that the bundled `JwtStrategy` only checks `cnf.x5t#S256`, not `cnf.jkt`, so a deployment relying solely on it shouldn't assume DPoP protection there yet.

Test plan:
- New: a wrapper `OpStore` reproducing the exact silent-jkt-drop bug proves the fail-closed guard fires.
- New: SQLite/Postgres TTL round-up regression tests; MySQL's reclaim test re-verified stable across repeated runs.
- New: 4 unit tests per adapter for header parsing (absent/one/duplicate/non-ASCII).
- `cargo test -p authkestra-op --lib`: 174 passed (up from 173).
- `cargo test -p authkestra-engine --lib --features memory,redis,sql-sqlite,sql-postgres,sql-mysql`: 21 `insert_if_absent` tests passed, including real Redis/Postgres/MySQL via testcontainers.
- `cargo test -p authkestra-axum --lib --features op`, `cargo test -p authkestra-actix --lib --features op`: header-parsing tests passing.
- `cargo doc -p authkestra-op --no-deps`: no new warnings.
- `cargo test --workspace --all-features`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check`: all clean.

marcjazz:
Follow-up tracked as #287 for the real ALTER-based migration (persisting `jkt`, and unblocking `token_endpoint_auth_method`/`jwks` on the same fix).

PR #288, opened 2026-08-30
feat(resource): DPoP (RFC 9449) key-binding verification in JwtStrategy

marcjazz:
## Summary

Phase C of #274. Phases A (#277) and B (#286) built and wired DPoP proof verification into `authkestra-op`'s `/token` endpoint: an OP can verify a client's proof and stamp `cnf.jkt` onto the tokens it issues. Nothing on the resource-server side checked that claim — a DPoP-bound access token was accepted by `authkestra-resource`'s `JwtStrategy` as an ordinary bearer token, with no proof-of-possession check at all. This PR closes that gap.

Targets `next`, not `main`, per the same release-timing reasoning as every other PR in this series.

## What's here

- **`authkestra_engine::token::dpop::verify_dpop_proof`'s `expected_htu` is now `Option<&str>`** instead of `&str`. An authorization server has exactly one, statically-known `/token` endpoint (still `Some(...)` at that call site), but a resource server protects many routes, commonly behind a reverse proxy or load balancer where the exact absolute URL a client used isn't always reconstructible from the request alone. `None` skips just that one check; `htm`, freshness, and (when supplied) `ath` are still verified — mirrors how `expected_ath` already models "not applicable here" as `None`.
- **New `authkestra_resource::dpop` module**: a `DpopReplayStore` trait with a blanket impl over any `KvStore + AtomicInsert` backend (Memory/Redis/Postgres/Sqlite/MySQL all get it for free), and a fail-closed `NoDpopReplayStore` default — deliberately independent from `authkestra-op`'s own replay guard, since an authorization server and a resource server are commonly separate deployments, each needing its own record of which proof `jti`s it has seen.
- **`ValidationConfig` gains `require_dpop: bool`** (off by default, mirroring `require_cert_binding`'s exact shape and rationale): when on, a token carrying `cnf.jkt` must be redeemed with a DPoP proof that verifies, binds to *this* access token via `ath`, carries the same key, and hasn't been seen before. A token with no `cnf.jkt` is untouched regardless of this setting.
- **`JwtStrategy` gains a `with_dpop_replay_store` builder** (mirrors `CompositeOpStore::with_dpop_replay_store` on the OP side) and a new `verify_dpop_binding` check in `authenticate()`, sharing the single `cnf` re-decode already done for certificate binding rather than decoding twice.
- **Both HTTP adapters' `extract_dpop_header` hardening** (reject a duplicate or non-ASCII `DPoP` header outright — from the `/token` endpoint's own review round) is mirrored here for the resource-server request path.
- **`with_dpop_support()`'s discovery doc comment updated**: the resource-server gap it flagged is now closeable via `require_dpop`/`with_dpop_replay_store`, but remains its own deployment-side opt-in — advertising DPoP support says nothing about whether a given resource server has actually turned enforcement on.

## Deliberate scope boundary

`htm`/`htu` verification at the resource server is out of scope here, documented on `ValidationConfig::require_dpop` — same reverse-proxy reasoning as the `expected_htu: None` change above. Key binding + `ath` + `jti` replay together still mean a captured proof cannot be reused for a different token, or reused at all; only "replay against a different route on the same server" is not caught. Flagging explicitly rather than silently shipping partial.

## Test plan

- [x] New `authkestra-resource` tests: valid proof accepted and its `jti` recorded; missing header, wrong key, wrong `ath`, wrong method, and replay all rejected; `NoDpopReplayStore` fails closed; a token with no `cnf.jkt` is unaffected regardless of `require_dpop`.
- [x] New `extract_dpop_header` unit tests (absent/one/duplicate/non-ASCII), mirroring the `/token` endpoint's own adapter tests.
- [x] New `authkestra-engine` test proving `expected_htu: None` skips the check; all 21 existing `verify_dpop_proof` call sites (engine tests) updated to `Some(...)` with zero behavior change.
- [x] `cargo test -p authkestra-resource --lib`: 25 passed (up from 13).
- [x] `cargo test -p authkestra-engine --lib token::dpop`: 23 passed (up from 22).
- [x] `cargo test -p authkestra-op --lib`: 174 passed, unchanged.
- [x] `cargo test --workspace --all-features`: clean.
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`: clean.
- [x] `cargo fmt --all -- --check`: clean.
- [x] `cargo doc -p authkestra-resource -p authkestra-op -p authkestra-engine --no-deps`: no new warnings.

## Follow-up

- `htm`/`htu` verification at the resource server, if a deployment topology allows reliably reconstructing the request's absolute URL (see the scope boundary above).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -674,6 +674,8 @@ dependencies = [
  "async-trait",
  "authkestra-engine",
  "base64 0.22.1",
+ "chrono",
+ "ed25519-dalek",
  "http 1.5.0",
  "jsonwebtoken",
  "rand 0.8.7",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -686,6 +686,7 @@ dependencies = [
  "thiserror 2.0.19",
  "tokio",
  "tracing",
+ "url",
  "wiremock",
 ]
 
--- a/crates/authkestra-engine/src/token/dpop.rs
+++ b/crates/authkestra-engine/src/token/dpop.rs
@@ -129,10 +129,23 @@ struct DpopClaims {
 /// Verifies a DPoP proof JWT (RFC 9449 §4.3).
 ///
 /// `expected_htm` is compared case-insensitively (§4.2: "the value of the
-/// HTTP method"). `expected_htu` and the proof's own `htu` are both
-/// stripped of query and fragment before comparison, per §4.3 step 6, using
-/// `url::Url` for canonicalization — pass in the request's URI as received,
-/// not pre-stripped.
+/// HTTP method") and always checked — a caller always knows the current
+/// request's method unambiguously, so there's no case where skipping it
+/// would be appropriate.
+///
+/// `expected_htu`, when `Some`, is compared against the proof's own `htu`
+/// with both sides stripped of query and fragment first, per §4.3 step 6,
+/// using `url::Url` for canonicalization — pass in the request's URI as
+/// received, not pre-stripped. `None` skips the `htu` check entirely: a
+/// resource server protecting many routes behind a reverse proxy or load
+/// balancer often cannot reliably reconstruct the exact absolute URL the
+/// client used (scheme and authority depend on the TLS-termination point,
+/// which isn't always visible to application code), unlike an
+/// authorization server's `/token` endpoint, which has exactly one, known,
+/// statically configured absolute URL. Skipping `htu` still leaves `jti`
+/// replay tracking and (when the caller supplies it) `ath` binding to
+/// prevent a captured proof from being reused — see the caller's own
+/// threat-model notes for what's traded away by not also checking `htu`.
 ///
 /// `expected_ath` should be `Some(base64url(SHA256(access_token)))` when
 /// verifying a proof presented alongside an access token (the resource
@@ -149,7 +162,7 @@ struct DpopClaims {
 pub fn verify_dpop_proof(
     compact_jws: &str,
     expected_htm: &str,
-    expected_htu: &str,
+    expected_htu: Option<&str>,
     expected_ath: Option<&str>,
     max_age: chrono::Duration,
 ) -> Result<VerifiedDpopProof, DpopError> {
@@ -258,8 +271,10 @@ pub fn verify_dpop_proof(
     if !claims.htm.eq_ignore_ascii_case(expected_htm) {
         return Err(DpopError::WrongHtm);
     }
-    if canonicalize_htu(&claims.htu) != canonicalize_htu(expected_htu) {
-        return Err(DpopError::WrongHtu);
+    if let Some(expected_htu) = expected_htu {
+        if canonicalize_htu(&claims.htu) != canonicalize_htu(expected_htu) {
+            return Err(DpopError::WrongHtu);
+        }
     }
 
     let now = chrono::Utc::now().timestamp();
@@ -431,7 +446,7 @@ mod tests {
         let verified = verify_dpop_proof(
             &proof,
             "POST",
-            "https://as.example.com/token",
+            Some("https://as.example.com/token"),
             None,
             chrono::Duration::seconds(60),
         )
@@ -447,7 +462,7 @@ mod tests {
         verify_dpop_proof(
             &proof,
             "post",
-            "https://as.example.com/token",
+            Some("https://as.example.com/token"),
             None,
             chrono::Duration::seconds(60),
         )
@@ -460,7 +475,7 @@ mod tests {
         let err = verify_dpop_proof(
             &proof,
             "GET",
-            "https://as.example.com/token",
+            Some("https://as.example.com/token"),
             None,
             chrono::Duration::seconds(60),
         )
@@ -474,7 +489,7 @@ mod tests {
         verify_dpop_proof(
             &proof,
             "POST",
-            "https://as.example.com/token?foo=bar#frag",
+            Some("https://as.example.com/token?foo=bar#frag"),
             None,
             chrono::Duration::seconds(60),
         )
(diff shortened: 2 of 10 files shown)


marcjazz:
All three confirmed and fixed in 68bd1c7 — thanks, the HIGH in particular was a real embarrassment: the feature genuinely couldn't be exercised by a spec-compliant client.

**HIGH (unreachable via the DPoP scheme), confirmed and fixed.** You're right — `authenticate` only ever recognized `Authorization: Bearer <token>`, so a real DPoP client (which RFC 9449 §7.1 requires present the token as `DPoP <token>`) fell straight through to `Ok(None)` before the new code could ever run. Added scheme-aware extraction local to this crate, and — since I was there — implemented the other half of §7.1 the reviewer's report implied but didn't need to spell out: a `cnf.jkt`-bound token presented via `Bearer` is now rejected outright, even if a genuinely valid `DPoP` proof header is also attached. Accepting either scheme for a bound token would let a client (or an attacker holding a stolen token) simply pick the one that doesn't require proof-of-possession.

**TTL anchor regression, confirmed and fixed.** Exactly the bug closed on the OP side two rounds ago, just never carried into this crate's copy — anchored to the proof's own `iat` now, matching that fix precisely (same 1s boundary buffer, same reasoning).

**Replay-store outage silently degrading, confirmed and fixed.** Added `ValidationError::DpopReplayUnavailable`, distinct from `InvalidToken`, so a store failure (or the fail-closed `NoDpopReplayStore` default) now propagates as `Err(AuthError)` instead of `Ok(None)` — matching the JWKS-resolution path's existing handling for the identical reason. This is a breaking addition to a non-`#[non_exhaustive]` public enum; `authkestra-oidc`'s exhaustive match needed (and got) a new arm.

Per your note that "no test drives `require_dpop` through `authenticate`" — added four true end-to-end integration tests in `tests/jwt_test.rs` that do exactly that (valid proof via `DPoP` scheme accepted; the same token via `Bearer` rejected even with a valid proof attached; `require_dpop=false` stays backward-compatible; replay-store outage surfaces as `Err`). Mutation-tested all three fixes by temporarily reverting each and confirming its test fails, then restoring and confirming green.

Test plan:
- `cargo test -p authkestra-resource`: 25 unit + 25 integration passed (up from 25 + 21).
- `cargo test -p authkestra-oidc --lib`: unchanged.
- `cargo test --workspace --all-features`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check`: all clean.
- `cargo doc -p authkestra-resource -p authkestra-oidc --no-deps`: no new warnings.

marcjazz:
All six confirmed and fixed in a29f8d8 — thanks for the careful re-review, and for verifying the previous round's fixes rather than just taking my word for them.

**1. `htu` opt-in (medium), confirmed and fixed.** You're right that leaving it unconditionally unchecked left a real cross-service forwarding gap. Added `ValidationConfig::dpop_resource_origin` (opt-in, `None`/off by default): when a deployment sets its own scheme+host, `authenticate` reconstructs the request's absolute URL from that origin + the request's own path and enforces `htu` against it — closing the gap for anyone who can state their own origin reliably, without forcing it onto deployments that can't (the reverse-proxy case the original doc comment described).

**2. Doc/code mismatch (low), confirmed and fixed.** `htm` was always checked; only `htu` was ever conditional. Doc rewritten to say so, and to describe the new opt-in.

**3. `memory` feature in production deps (low), confirmed and fixed.** Moved to a dev-dependency-only declaration. Verified with `cargo tree -p authkestra-resource -e normal` that it no longer activates for a production build.

**4. Non-`#[non_exhaustive]` breaking enum (medium), confirmed and fixed.** Fair catch — this file documents the exact convention for exactly this reason and I missed applying it to the one enum that needed it most. Added `#[non_exhaustive]` to `ValidationError` and gave `authkestra-oidc`'s match a wildcard arm, so the next new variant doesn't repeat this.

**5. Case-sensitive scheme matching (low), confirmed and fixed.** `extract_presented_token` now compares the scheme case-insensitively (RFC 9110 §11.1) for both `DPoP` and `Bearer`.

**6. DPoP-scheme presentation of an unbound token (low), confirmed and fixed.** Added the missing direction: a request claiming the `DPoP` scheme for a token with no `cnf.jkt` is now refused. Also took your note on the duplicated predicate and extracted `dpop_bound_jkt` as the single "is this token bound" check shared by `authenticate` and `verify_dpop_binding`, so they can't drift apart again.

Test plan:
- New end-to-end tests for the `htu` opt-in (accept-when-matching, reject-when-mismatched) and the DPoP-scheme/unbound-token case.
- New `presented_token_tests` unit module (5 tests) covering case-insensitive scheme recognition, including the exact `Authorization: dpop ...` case from the report.
- Mutation-tested all three new checks by reverting each individually and confirming its test fails, then restoring.
- `cargo test -p authkestra-resource`: 30 unit + 28 integration passed (up from 25 + 25).
- `cargo test -p authkestra-oidc`: 19 passed, unchanged.
- `cargo test --workspace --all-features`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check`: all clean.
- `cargo doc -p authkestra-resource -p authkestra-oidc --no-deps`: no new warnings.

marcjazz:
Confirmed and fixed in f1dc01c.

You're right — both a scheme-less origin and a trailing-slash origin would silently fail *every* DPoP request once `require_dpop` is on, with only a `tracing::warn!` per request and nothing pointing at the actual cause. `ValidationConfigBuilder::build` now asserts on both shapes at startup, mirroring the existing missing-`jwks_url` panic it already documents — a static config mistake should fail loudly once at boot, not silently 401 every request in production.

Test plan:
- New: a well-formed origin round-trips correctly; a scheme-less origin panics with a clear message; a trailing-slash origin panics with a clear message.
- `cargo test -p authkestra-resource`: 33 unit + 28 integration passed (up from 30 + 28).
- `cargo test --workspace --all-features`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check`: all clean.
- `cargo doc -p authkestra-resource --no-deps`: no warnings.

marcjazz:
Both residual nits were cheap enough to close properly rather than leave as a deliberate gap — pushed in 29b53fe.

Switched the `dpop_resource_origin` shape check from hand-rolled string prefix/suffix checks to real `url::Url::parse`: this naturally normalizes scheme case (closing the `HTTPS://` false-panic) and exposes `.query()`/`.fragment()` directly, so a query or fragment on the origin is now rejected at startup instead of silently producing an `htu` that could never match. The trailing-slash check stays on the raw string on purpose — `Url` treats a trailing slash as equivalent to no path, but the runtime code concatenates the raw configured string, so only a raw-string check actually catches that specific bug.

Test plan:
- New: origin-with-query panics, origin-with-fragment panics, uppercase scheme accepted, non-http(s) scheme panics.
- `cargo test -p authkestra-resource`: 37 unit + 28 integration passed (up from 33 + 28).
- `cargo test --workspace --all-features`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check`: all clean.
- `cargo doc -p authkestra-resource --no-deps`: no warnings.

Issue #289, opened 2026-08-30
[Epic] Rework storage architecture: executor-composable traits, satellite SQL crate, conformance test suite


marcjazz:
## Context

Came up while scoping #287 (the SQL migration follow-up from #274/DPoP). The narrow question was "how do we add columns to an existing deployment's table safely" — but that surfaced a bigger architectural question about where SQL storage belongs and how the storage traits are shaped, which is genuinely a separate, larger piece of work. #287 stays scoped to the small migration-mechanism fix; this issue tracks the bigger rework.

## The problem with the current shape

- `authkestra-op` (and `authkestra-engine`) own a SQL-backed store implementation in core, which means core owns the sqlx feature matrix (`sql-postgres`/`sql-mysql`/`sql-sqlite`), a second connection pool concept, and a migration system that competes with whatever the host application already uses for its own schema.
- `OpStore`/`ClientStore`/`RefreshTokenStore`/etc. methods take `&self` and manage their own storage calls internally. A host application that wants "create user + create org + create refresh token" as one atomic operation has no way to get that — the trait gives it no way to pass in its own transaction/executor.
- An auth store is not a neutral CRUD layer: token hashing, constant-time lookups, session rotation on refresh, expiry cleanup, concurrent-refresh races (see the MySQL deadlock and Redis/SQL TTL-rounding bugs found and fixed during the DPoP work — authkestra#277's review rounds) are all easy to get subtly wrong. If every third-party backend (a different ORM, a different database) has to reimplement this from scratch with no shared test coverage, the long tail of subtly-insecure implementations gets attributed to authkestra regardless of whose code it actually is.
- Documentation-only integration examples (markdown snippets) rot silently — they're not compiled or run in CI, so they can go stale without anyone noticing until a user hits it.

## Proposed direction

1. **Executor/transaction-composable traits.** Rework `ClientStore`/`AuthorizationCodeStore`/`RefreshTokenStore`/`DeviceCodeStore` (and whatever else needs it) so a caller can supply its own executor/transaction, letting a host application compose auth-store operations into its own atomic units of work. This is a breaking change to every existing trait signature and every existing implementor (Memory, Redis, SQL) — needs its own design pass on exactly what the generic bound / associated-type shape looks like before implementation starts.
[...]

PR #290, opened 2026-08-30
fix(op): real ALTER-based migrations for SqlxOpStore via sqlx::migrate!

### LINKED ISSUE #287: op(sqlx): idempotent ALTER-based migrations so SqlxOpStore can persist jkt, token_endpoint_auth_method, jwks
## Context

Follow-up from #274 (DPoP), specifically from PR #286's review. `SqlxOpStore`'s `migrate()` is `CREATE TABLE IF NOT EXISTS` for every backend (Postgres/SQLite/MySQL), which is a no-op against a table that already exists. That's fine for creating fresh schema, but it means **no column can ever be added to an existing deployment's table** without breaking it outright (`unknown column` / `column does not exist` on every INSERT/SELECT that names it) — there's no migration path from "table exists without column X" to "table exists with column X".

This limitation currently blocks three unrelated fields, all stubbed out with the identical `None`-and-a-comment pattern because of it:

- `RefreshToken.jkt` (RFC 9449 §5 DPoP refresh-token key continuity) — added in #286, but never persisted by `SqlxOpStore`'s `RefreshTokenStore` impl. A fail-closed guard (also from #286) means this doesn't silently fail — SQL-backed deployments now get a loud `server_error` if they try to combine DPoP with `offline_access` — but they still can't actually use the combination.
- `ClientRegistration.token_endpoint_auth_method` — blocks `private_key_jwt` for SQL-backed clients.
- `ClientRegistration.jwks` — same.

See the comments on `sqlx_store.rs`'s `get_token`/`find_client` for the exact reasoning already documented at each site.

## History: `sqlx::migrate!` tried and reverted

The first attempt at this ticket (0839ed3, PR #290) adopted `sqlx::migrate!`, sqlx's own built-in versioned-migration runner. Review of that PR found a real regression: `sqlx::migrate!`'s bookkeeping lives in one database-global, unqualified `_sqlx_migrations` table with no supported way in sqlx 0.8 to rename or namespace it. `SqlxOpStore` is embedded into a *host application's own* connection pool, and a host that also runs `sqlx::migrate!` for its own schema against that same pool — an entirely ordinary setup — collides on that shared table the moment either side's migration lands on a version number the other already used, with `Migrate::run` then refusing to apply *either* migration set. The old `CREATE TABLE IF NOT EXISTS`-only approach never had this failure mode, because it kept no bookkeeping of its own at all — so this was a genuine regression, not a wash.

Fixed in 9c96bfa by reverting to bookkeeping-free schema setup while still gaining the ability to add columns to an existing deployment: Postgres uses its native `ALTER TABLE ... ADD COLUMN IF NOT EXISTS` (9.6+); SQLite and MySQL introspect first (`pragma_table_info` / `information_schema.columns`) since neither backend supports that syntax directly. No `.sql` migration files, no shared bookkeeping table, no dependency on the `migrate` sqlx feature.

## Scope (as actually implemented)

- Each backend's `migrate()` runs the original `CREATE TABLE IF NOT EXISTS` schema plus introspection-guarded (SQLite/MySQL) or native `IF NOT EXISTS` (Postgres) `ALTER TABLE` statements adding `jkt` to `oauth_refresh_tokens` and `token_endpoint_auth_method`/`jwks` to `oauth_clients`.
- The new columns are wired into `sqlx_store.rs`'s `RefreshTokenStore`/`ClientStore` impls, reverting the `None`-stub comments added in #286.
- A decode error on `token_endpoint_auth_method`/`jwks`/`jkt` is now propagated as `OpError::Storage` rather than silently discarded (a fail-open bug found in the same PR #290 review — see that PR's discussion for the exploit path).
- Test plan: real Postgres/SQLite/MySQL integration tests via testcontainers proving the migration is safe to run against **both** a fresh install and a database that already has the pre-migration schema, plus a SQLite-specific regression test proving `migrate()` does not collide with a host application's own independent `sqlx::migrate!` run against the same pool.

## Non-goals

- No changes to the `CompositeOpStore`/Memory/Redis-backed paths, which already persist `jkt` correctly.
- No changes to trait shapes, crate boundaries, or where `SqlxOpStore` lives — that's #289.

marcjazz:
## Summary

Closes #287. `SqlxOpStore`'s `migrate()` was raw `CREATE TABLE IF NOT EXISTS` execution — fine for a fresh install, a no-op against a table that already exists, so it could only ever create schema, never evolve it. That blocked three fields, each stubbed to `None` with an identical comment: `RefreshToken.jkt` (RFC 9449 DPoP refresh-token key continuity, added in #286), and `ClientRegistration.token_endpoint_auth_method`/`jwks` (RFC 7523 `private_key_jwt`).

Rather than hand-rolling a per-backend "check column exists, then conditionally `ALTER`" helper (the originally-scoped approach — see #287's revised body), adopted sqlx's own built-in migration runner: `sqlx::migrate!`, already available since this crate depends on `sqlx` directly, zero new dependencies. Ordered, checksummed `.sql` files under `migrations/{postgres,sqlite,mysql}/`, tracked in each database's own `_sqlx_migrations` table so every migration runs exactly once, ever, per database — the primitive that was actually missing, and one that eliminates the need for defensive `IF NOT EXISTS`/introspection on the `ALTER` statements themselves.

Targets `next`, matching the branch this whole DPoP-adjacent thread of work has staged on.

## What's here

- `migrations/*/1_initial.sql`: byte-for-byte the previous `CREATE TABLE IF NOT EXISTS` statements, so a fresh install is unaffected.
- `migrations/*/2_add_dpop_and_client_auth_columns.sql`: plain `ALTER TABLE ADD COLUMN` for `jkt`, `token_endpoint_auth_method`, and `jwks` — no guards needed, since sqlx's own tracking already guarantees this file runs once.
- Wired the three columns into `store_token`/`get_token`/`consume_token` and `find_client`, removing the `None`-stub comments from #286.

## Test plan

- [x] New per-backend **upgrade** test: creates the pre-#287 schema directly (bypassing the new `migrate()` entirely, simulating an existing deployment), inserts data under it, then runs the new `migrate()` and confirms it succeeds, pre-existing data survives, and the new columns are genuinely usable afterwards. This is the exact scenario `CREATE TABLE IF NOT EXISTS` alone could never handle — there was no prior testcontainers coverage of `RefreshTokenStore`'s SQL persistence at all.
- [x] Mutation-tested the SQLite version: temporarily reverted `migrate()` to a no-op, confirmed the upgrade test fails, restored, confirmed green.
- [x] New per-backend **fresh install** test proving `jkt` and `token_endpoint_auth_method`/`jwks` round-trip correctly end to end.
- [x] Verified via `cargo package --list` that the new `migrations/` directory ships in the published crate (`sqlx::migrate!` reads these files at compile time, so they must be present in any build, not just a git checkout).
- [x] `cargo test -p authkestra-op --features sqlx-postgres,sqlx-sqlite,sqlx-mysql --lib`: 186 passed (up from 174), including 12 in `sqlx_store` (up from 6) with real Postgres/MySQL/SQLite via testcontainers.
- [x] `cargo build -p authkestra-op` (default features, no SQL backends): clean — confirms the `sqlx::migrate!` invocations stay properly `cfg`-gated per backend.
- [x] `cargo test --workspace --all-features`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check`: all clean.
- [x] `cargo doc -p authkestra-op --no-deps --all-features`: no new warnings.

## Follow-up

Tracked separately in #289: the bigger question of whether SQL storage belongs in core at all (executor-composable traits, extracting `SqlxOpStore` to a satellite crate, a conformance test suite, real compiled ORM examples). Explicitly out of scope here.


marcjazz:
Confirmed both HIGHs and fixed in 9c96bfa.

**HIGH #1 (approach itself)** — reproduced exactly: with a host-app migration at version 1 applied first, `store.migrate()` failed with `Migrate(VersionMismatch(1))`; reversed, the host's own migrator broke instead. Verified against the vendored sqlx-core/sqlx-postgres 0.8.6 source that `_sqlx_migrations` is a hardcoded, unqualified table name with no supported way to rename or namespace it in sqlx 0.8 — so a dedicated schema/`search_path` wasn't actually available as an escape hatch either, since the bookkeeping table itself is global regardless of where the *data* tables live.

Went with introspection-guarded ALTERs, dropping `sqlx::migrate!` entirely: Postgres uses native `ADD COLUMN IF NOT EXISTS` (9.6+); SQLite and MySQL introspect first (`pragma_table_info` / `information_schema.columns`) since neither has that syntax. This restores the bookkeeping-free property the original `CREATE TABLE IF NOT EXISTS` code had — no shared state, no collision surface — while still being able to add a column to an existing deployment. `migrations/` and the `migrate` sqlx feature are removed.

New regression test `test_sqlite_migrate_does_not_collide_with_a_host_apps_own_sqlx_migrate` runs a real `sqlx::migrate!` against a fixture directory in both orderings against `SqlxOpStore::migrate` sharing one pool. Mutation-tested: temporarily restored `sqlx::migrate!` on the SQLite backend and reproduced the exact `VersionMismatch(1)` failure against this test before restoring the fix.

**HIGH #2 (fail-open decode)** — confirmed by reading `authenticate_client`: `(None, Cred::NoCredential)` with `client_secret_hash: None` (a legitimate shape for a private_key_jwt-only client) returns `Ok(())`, so a decode failure on `token_endpoint_auth_method` really does authenticate with zero credentials. `.ok().flatten()` → `.map_err(|_| OpError::Storage)?` on `token_endpoint_auth_method` and `jwks` in `find_client`; same fix for `jkt`'s `.ok()` in `get_token` and all three per-backend `consume_token` impls, since that one drops the RFC 9449 §5 continuity check instead. `try_get::<Option<Json<T>>, _>` still reports genuine SQL NULL as `Ok(None)` — only a non-NULL value that fails to decode now propagates as an error, matching what you flagged.

New test `test_sqlite_find_client_rejects_an_undecodable_token_endpoint_auth_method`: a client row with an enum-incompatible `token_endpoint_auth_method` (`client_secret_jwt`) now makes `find_client` return `Err(OpError::Storage)`. Mutation-tested: temporarily restored `.ok().flatten()` and confirmed this test fails with the exact fail-open `Ok(Some(ClientRegistration { token_endpoint_auth_method: None, .. }))` result before restoring the fix.

Test plan run: `cargo test -p authkestra-op --features sqlx-postgres,sqlx-sqlite,sqlx-mysql --lib` (188 passed, including all 12 pre-existing fresh-install/migration-upgrade tests unchanged), `cargo test --workspace --all-features`, `cargo clippy --workspace --all-targets --all-features -- -D warnings` (clean), `cargo fmt --all -- --check` (clean), `cargo doc -p authkestra-op --no-deps --all-features` (no new warnings), `cargo package --list -p authkestra-op` (no `migrations/` in the package).

marcjazz:
@/tmp/claude-1000/-home-marco-Projects-personnal-authkestra/09fdc300-a4f4-424f-a7ef-a7222a4a23fb/scratchpad/comment_body_fixed.md

marcjazz:
#4 fixed in 1f78aa0 — confirmed you're right, this was missed entirely.

Checked: `find_client`'s SELECT and `store_token`'s INSERT both name `token_endpoint_auth_method`/`jwks`/`jkt` unconditionally (not gated on whether `migrate()` ran), and the SQL store doc explicitly tells readers they can "skip `migrate()` and create the tables yourself" with zero mention that these three columns are now a hard requirement either way. A deployment following that documented path breaks on every token request immediately post-upgrade, with nothing anywhere explaining why or what to add. `sqlx_store.rs` itself is deliberately untouched — making the columns conditionally-optional in code would reintroduce exactly the silent-degradation shape the HIGH #2 fix closed, so this is a docs fix, not a code one: added an "Upgrading a hand-managed schema" section to the SQL store doc with the exact `ALTER TABLE` statements per backend and a note that all three are nullable/additive.

On the residual note for #1: agreed, and worth being precise about it rather than letting it sound resolved. The pre-consumption comment's invariant ("a request that fails never has a side effect on a token it had no right to consume") is fully true for the checks it actually covers — client_id, expiry, DPoP continuity, all validated against the read-only candidate before `consume_token` runs. It was never intended to cover the confirmation probe *after* consumption, and doesn't claim to. What changed is narrower: a transient read failure on that probe no longer destroys a session for a write that actually succeeded. A deployment whose store deterministically drops `jkt` still loses the presented token on every rotation attempt — but that's a fixed, discoverable misconfiguration (the existing `a_backend_that_silently_drops_jkt_fails_the_request_instead_of_lying` test is deliberately about exactly this case staying loud), not the random, session-destroying transient failure that was the actual harm. No further code change proposed here since none was asked for.

PR #318, opened 2026-09-04
Generate an OAuth2 nonce only when the provider validates one

### LINKED ISSUE #317: OAuth2Flow sets a nonce unconditionally, so plain OAuth2 providers can never complete a login
## Summary

`OAuth2Flow::initiate_login` sets an OIDC nonce unconditionally, and `finalize_login` then requires the returned `Identity` to carry a matching one. The shipped providers cannot satisfy that, so **every plain-OAuth2 round trip fails** — GitHub, Google and Discord alike.

Found while building the playground against `authkestra 0.8.0` (crates.io).

## The mechanism

`crates/authkestra-engine/src/flow/oauth2.rs:128` — a nonce is always generated:

```rust
let nonce = Some(uuid::Uuid::new_v4().to_string());
```

`oauth2.rs:192` — and then demanded back:

```rust
if let Some(expected_nonce) = expected_state.nonce.as_deref() {
    if identity.attributes.get("nonce").map(|s| s.as_str()) != Some(expected_nonce) {
        tracing::error!("nonce mismatch or missing in identity attributes");
        return Err(AuthError::Token("Nonce mismatch".to_string()));
    }
}
```

But the providers generated by `define_oauth_provider!` ignore it — `crates/authkestra-providers/src/macros.rs:129` takes `_nonce: Option<&str>` (unused), and the identity is built with `attributes: std::collections::HashMap::new()`.

So `expected_state.nonce` is always `Some`, `identity.attributes["nonce"]` is always absent, and the comparison always fails.

A nonce is an ID-token concept. Plain OAuth2 has no ID token, so there is nothing for it to bind to.

## Observed

Real server logs from three separate attempts:

```
ERROR callback{provider=google}:finalize_login{provider_id=google}:
      authkestra_engine::flow::oauth2: nonce mismatch or missing in identity attributes

ERROR callback{provider=discord}:finalize_login{provider_id=discord}:
      authkestra_engine::flow::oauth2: nonce mismatch or missing in identity attributes
```

Everything before this works: the authorization redirect carries `state` and PKCE (`code_challenge_method=S256`), the provider redirects back, the encrypted state cookie decrypts, and the code exchange itself succeeds. The failure is purely the nonce assertion afterwards.

## Reproducing

Any `OAuth2Flow` over a `define_oauth_provider!` provider, driven through `authkestra_axum::helpers::initiate_oauth_login` + `handle_oauth_callback_erased`, with real credentials. It fails after a successful exchange.

## Suggested fix

The check is right; setting the nonce is what is wrong. Some options, roughly in order of preference:

1. **Don't set a nonce in the generic `OAuth2Flow`.** It is not an OIDC flow. Let `authkestra-oidc` set one where an ID token is actually validated. Smallest change, and it makes the invariant "a nonce exists only if something checks it".
2. Make it opt-in — a builder flag or a `Provider` capability like `validates_nonce()` — so a provider that does validate ID tokens can request one.
3. Have each provider echo the nonce it was passed into `attributes`. This satisfies the check but is security theatre: the value never leaves the server, so comparing it proves nothing.

Happy to send a PR for (1) or (2) if you tell me which you would prefer — (2) is more work but keeps the door open for providers that do validate an ID token.

## Workaround meanwhile

The playground clears the nonce from the state cookie between the login redirect and the callback. `state` and the PKCE verifier are untouched, so CSRF and code-interception protection are unaffected — there is simply no ID token for a nonce to bind to. Tracked to be removed once this is fixed upstream.

marcjazz:
Fixes #317.

## The bug

`OAuth2Flow::initiate_login` set a nonce unconditionally, and `finalize_login` then required a matching one back in `identity.attributes`. A nonce binds an **ID token** to the authorization request that asked for it — plain OAuth2 has no ID token, so the providers generated by `define_oauth_provider!` take `_nonce` as an unused parameter and build their identity with an empty attribute map.

Net effect: **no plain-OAuth2 login could complete.** All three shipped providers failed *after* a successful code exchange:

```
ERROR callback{provider=google}:finalize_login{provider_id=google}:
      nonce mismatch or missing in identity attributes
ERROR callback{provider=discord}:finalize_login{provider_id=discord}:
      nonce mismatch or missing in identity attributes
```

## The fix

`OAuthProvider` gains an opt-in capability, defaulting to the safe value:

```rust
fn validates_nonce(&self) -> bool { false }
```

`OAuth2Flow::initiate_login` generates a nonce only when it is `true`. `OidcProvider` overrides it — it sends the nonce in the authorization URL, verifies it against the ID token's `nonce` claim, and surfaces it in `Identity::attributes`.

Additive and backward compatible: a default trait method, so existing implementors compile unchanged and keep the conservative behaviour.

## Why not simply drop the nonce from `OAuth2Flow`

That was my first suggestion in #317 and it is **wrong**. `OidcProvider::get_authorization_url` appends `&nonce=` only when given one (`crates/authkestra-oidc/src/provider.rs:425`), so removing the generation would have stopped a nonce ever reaching the provider — silently removing ID-token replay protection from OIDC while fixing plain OAuth2.

Worth noting the existing comment at `provider.rs:513` documents this coupling: the OIDC provider echoes the nonce into attributes specifically to satisfy the generic re-check, referencing `#225 Defect 5`. That patch fixed the OIDC side of the interaction; the plain-provider side was never addressed.

## Tests

Three added in `crates/authkestra-engine/src/flow/oauth2.rs`:

- `no_nonce_is_generated_for_a_provider_that_cannot_validate_one`
- `a_nonce_is_generated_for_a_provider_that_validates_one` — guards the OIDC regression
- `a_plain_provider_can_complete_a_login` — the end-to-end shape of the bug; this one failed before the change

## Verified

```
cargo fmt --all -- --check                                  clean
cargo clippy --workspace --all-features -- -D warnings      clean
cargo test --workspace --all-features                       all green (149 in authkestra-engine)
```

Found while building the playground against `0.8.0` from crates.io. The playground currently carries a workaround that strips the nonce from the state cookie between the redirect and the callback; I will remove it once this lands.

### DIFF
--- a/crates/authkestra-engine/src/auth/mod.rs
+++ b/crates/authkestra-engine/src/auth/mod.rs
@@ -182,6 +182,25 @@ pub trait OAuthProvider: Provider {
     /// Get the provider identifier.
     fn provider_id(&self) -> &str;
 
+    /// Whether this provider validates an OIDC nonce.
+    ///
+    /// A nonce binds an ID token to the authorization request that asked for
+    /// it. Plain OAuth2 has no ID token, so there is nothing for a nonce to
+    /// bind to and nothing that can echo it back — which is why this defaults
+    /// to `false`.
+    ///
+    /// [`crate::flow::OAuth2Flow`] generates a nonce only when this is `true`,
+    /// because its `finalize_login` requires a matching nonce in the returned
+    /// identity's attributes. Generating one for a provider that cannot return
+    /// it makes every login fail.
+    ///
+    /// OIDC providers override this: they send the nonce in the authorization
+    /// URL, verify it against the ID token's `nonce` claim, and surface it in
+    /// `Identity::attributes`.
+    fn validates_nonce(&self) -> bool {
+        false
+    }
+
     /// Helper to get the authorization URL.
     fn get_authorization_url(
         &self,
--- a/crates/authkestra-engine/src/auth/mod.rs
+++ b/crates/authkestra-engine/src/auth/mod.rs
@@ -182,21 +182,24 @@ pub trait OAuthProvider: Provider {
     /// Get the provider identifier.
     fn provider_id(&self) -> &str;
 
-    /// Whether this provider validates an OIDC nonce.
+    /// Whether this provider validates an OIDC nonce **and surfaces it in
+    /// [`crate::state::Identity::attributes`]**.
     ///
     /// A nonce binds an ID token to the authorization request that asked for
     /// it. Plain OAuth2 has no ID token, so there is nothing for a nonce to
-    /// bind to and nothing that can echo it back — which is why this defaults
-    /// to `false`.
+    /// bind to and nothing to surface — which is why this defaults to `false`.
     ///
-    /// [`crate::flow::OAuth2Flow`] generates a nonce only when this is `true`,
-    /// because its `finalize_login` requires a matching nonce in the returned
-    /// identity's attributes. Generating one for a provider that cannot return
-    /// it makes every login fail.
+    /// [`crate::flow::OAuth2Flow`] hands every provider a nonce regardless of
+    /// this flag. What the flag controls is whether `finalize_login`
+    /// *re-checks* it against `attributes["nonce"]` afterwards. So a provider
+    /// that forgets to override this loses only a redundant second check, never
+    /// the nonce itself — the reason the flag gates the check and not the
+    /// generation.
     ///
-    /// OIDC providers override this: they send the nonce in the authorization
-    /// URL, verify it against the ID token's `nonce` claim, and surface it in
-    /// `Identity::attributes`.
+    /// Return `true` only if both halves hold: the nonce is verified against
+    /// the ID token's `nonce` claim, **and** copied into
+    /// `Identity::attributes`. Verifying without copying would fail the
+    /// re-check. `authkestra-oidc`'s provider does both.
     fn validates_nonce(&self) -> bool {
         false
     }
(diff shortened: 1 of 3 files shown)


marcjazz:
Adopted — you're right, and the reasoning is better than what I had. Pushed.

**The gate moved from generation to the check** (`oauth2.rs`), so a nonce is generated and handed to every provider as before, and only the engine's re-check is conditional.

Two things I verified while making the change, one of which strengthens your case:

**Plain providers *do* put the nonce in the authorization URL.** `define_oauth_provider!` appends `&nonce=` whenever it's given one (`crates/authkestra-providers/src/macros.rs`), so 0.8.0 already sends a nonce to GitHub, Google and Discord today. That means my version wasn't merely riskier for out-of-tree providers — it would have *removed a query parameter from every plain-provider auth URL*, a wire-format change for no benefit. Yours leaves the wire exactly as shipped and touches only enforcement. Your "a nonce sent to a provider that ignores it is harmless" holds for all three in-tree providers.

**I sharpened the trait doc**, because gating the check makes the contract stricter than the name suggests: the re-check reads `identity.attributes["nonce"]`, so returning `true` requires *both* that the provider verifies the nonce against the ID token claim **and** that it copies it into `Identity::attributes`. Verifying without copying would now fail the re-check. `OidcProvider` does both (`provider.rs:501` and `:517`).

That raises a naming question I'll leave to you: `validates_nonce()` describes half of what it asserts. Something like `surfaces_validated_nonce()` would be precise but clunkier. I kept your name to avoid churn on an API you own — happy to rename if you'd prefer.

## Tests

Rewritten for the new semantics, since two of mine asserted the old behaviour:

- `a_plain_provider_can_complete_a_login` — the original regression
- `every_provider_still_receives_a_nonce` — the point of gating the check rather than generation
- `a_wrong_nonce_is_still_rejected` — proves gating didn't disable enforcement where a provider opts in
- `a_matching_nonce_is_accepted` — happy path

I also checked the gate is load-bearing by forcing it to `true`: `a_plain_provider_can_complete_a_login` then fails, which is exactly the reported bug.

```
cargo fmt --all -- --check                                clean
cargo clippy --workspace --all-features -- -D warnings    clean
cargo test --workspace --all-features                     60 targets green
```

Thanks for catching the out-of-tree case — that failure mode was invisible from where I was standing.

PR #322, opened 2026-09-05
fix(providers): make single-provider builds work (#321)

### LINKED ISSUE #321: authkestra-providers fails to compile with only the `github` or `google` feature
## Summary

`authkestra-providers` declares `urlencoding` as an optional dependency enabled **only** by the `discord` feature, but uses it in the macro that generates **every** provider. So depending on the crate with just `github`, or just `google`, fails to compile inside the dependency.

## Reproducing

```toml
[dependencies]
authkestra-providers = { version = "0.8.0", features = ["github"] }
```

```
error[E0433]: cannot find module or crate `urlencoding`
   --> authkestra-providers-0.8.0/src/macros.rs:108:36
    |
108 |     redirect_uri = urlencoding::encode(&self.redirect_uri),
    |                    ^^^^^^^^^^^ use of unresolved module or unlinked crate `urlencoding`

error[E0433]: cannot find module or crate `urlencoding`
   --> authkestra-providers-0.8.0/src/macros.rs:110:35
```

Same with `features = ["google"]`. Adding `discord` makes it compile.

## Why

`crates/authkestra-providers/Cargo.toml`:

```toml
[features]
default = ["rustls-aws-lc-rs"]
discord = ["urlencoding"]
github = []
google = []
```

`urlencoding` is behind `discord`, but `macros.rs` uses it unconditionally — it is in `define_oauth_provider!`, which generates GitHub and Google too.

The workspace itself never hits this: everything in-tree that uses the crate enables all three providers at once, so `urlencoding` is always on.

## Impact

The `github` and `google` features are individually unusable. Anyone wanting a single provider has to enable `discord` as well, and nothing says so — the error appears inside a dependency, which is a confusing place to land.

Found while generating starter projects: a project selecting only GitHub sign-in does not build.

## Suggested fix

`urlencoding` is used by all providers, so make it unconditional:

```toml
[dependencies]
urlencoding = "2"      # not optional

[features]
discord = []
```

Alternatively, if the intent is to keep it optional, gate the macro's use of it — but since every generated provider percent-encodes its redirect URI and scopes, unconditional looks right.

Either way a CI check that each provider feature builds on its own would catch this class of thing:

```sh
for f in github google discord; do
  cargo check -p authkestra-providers --no-default-features --features "$f,rustls-aws-lc-rs" || exit 1
done
```

## Workaround meanwhile

The playground's starter-kit generator adds `discord` to the feature list whenever any provider is selected, with a comment pointing here. It costs an unused provider in the dependency graph and nothing else.

marcjazz:
Fixes #321. **Patch-safe: nothing in the public API changes.**

`authkestra-providers` gated `urlencoding` behind the `discord` feature alone, but uses it in the macro that generates *every* provider. So `features = ["github"]` failed to compile inside the dependency:

```
error[E0433]: cannot find module or crate `urlencoding`
   --> authkestra-providers-0.8.0/src/macros.rs:108:36
```

`urlencoding` is now a plain dependency and `discord` an empty feature, matching what `authkestra-oidc` already does.

### Keeping it a patch

Two things had to be handled, both raised in review:

- Making the dependency non-optional also removes the **implicit feature** named `urlencoding` that the optional dependency created. Anyone who enabled it to work around #321 would hit `does not have the feature urlencoding` — a resolution error, not a compile error, so unworkable-around downstream. It is kept as a no-op.
- The `AxumError::NotFound` work originally in this PR has moved out. Adding a variant to a public enum that is neither `#[non_exhaustive]` nor unreachable is a major bump under Cargo's SemVer rules, and `AxumError` is `pub use`d at the crate root and serves as the extractor `Rejection` type. It will land in the next minor instead.

### The guard

A `provider-features` matrix builds each provider on its own. It fails on the parent commit and passes here.

It initially ran `cargo check` without `--all-targets`, which compiles only the lib — so it guarded one shape of the bug rather than the class. The provider tests had no feature gating, and

```sh
cargo check --all-targets -p authkestra-providers --no-default-features \
  --features github,rustls-aws-lc-rs
```

failed with unresolved imports of the `google` and `discord` modules while the job passed. The tests now declare `required-features` (the pattern `authkestra-axum` already uses) and the job builds `--all-targets`.

Verified load-bearing by removing `required-features` from one test and watching the job's exact command fail, then restoring it.

### Note for the release

Worth a changelog line: the `urlencoding` feature still exists but does nothing, and can be dropped at the next major.

### Why the workspace never saw this

Correcting what this section said earlier: I claimed `crates/authkestra/Cargo.toml:29` pinned the provider features on the facade's *dependency* line. It does not — line 29 is in `[dev-dependencies]`. The facade's real dependencies (13-19) are all `optional = true, default-features = false`, which is right.

The mechanism is dev-dependency feature unification instead. Any workspace test build compiles those dev-dependencies, so the single `authkestra-providers` in the graph gets `github`, `google` and `discord` all enabled, and `urlencoding` is always present. Combined with every CI job building `--all-features`, no build in the repo had ever resolved a single provider on its own.

The conclusion stands and so does the fix; the explanation was wrong.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,27 @@ jobs:
       - name: Run tests
         run: cargo test --workspace --all-features
 
+  provider-features:
+    # Every other job builds with --all-features, which is why a dependency
+    # gated behind one provider feature but used by all of them compiled
+    # cleanly here and broke for anyone selecting a single provider.
+    # See https://github.com/marcjazz/authkestra/issues/321.
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        feature: [github, google, discord]
+    steps:
+      - uses: actions/checkout@v7
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+      - name: Cache cargo registry and build
+        uses: swatinem/rust-cache@v2
+      - name: Build with one provider and nothing else
+        run: |
+          cargo check -p authkestra-providers --no-default-features \
+            --features ${{ matrix.feature }},rustls-aws-lc-rs
+
   lint:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,8 +47,12 @@ jobs:
       - name: Cache cargo registry and build
         uses: swatinem/rust-cache@v2
       - name: Build with one provider and nothing else
+        # --all-targets on purpose: without it this only compiles the lib, and
+        # the tests and examples — where a feature-gating slip is just as easy
+        # to make — are never built. Requires the provider tests to declare
+        # `required-features`, which they do.
         run: |
-          cargo check -p authkestra-providers --no-default-features \
+          cargo check --all-targets -p authkestra-providers --no-default-features \
             --features ${{ matrix.feature }},rustls-aws-lc-rs
 
   lint:
--- a/crates/authkestra-providers/Cargo.toml
+++ b/crates/authkestra-providers/Cargo.toml
@@ -15,6 +15,13 @@ default = ["rustls-aws-lc-rs"]
 github = []
 google = []
 discord = []
+# A no-op, kept deliberately. `urlencoding` used to be an optional dependency,
+# which gave this crate an implicit feature of the same name. Making it a real
+# dependency removes that feature, and a manifest asking for a feature that no
+# longer exists fails to resolve — an error a downstream crate cannot work
+# around. Anyone who reached for `features = ["urlencoding"]` to get past #321
+# keeps resolving.
+urlencoding = []
 
 # TLS backend, forwarded to authkestra-engine. See its documentation for the
 # `rustls-no-provider` contract.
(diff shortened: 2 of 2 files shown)


marcjazz:
Restructured per your review — this is now providers-only and ships as a **patch**.

**On the major version.** You're right to push back, and the fix was to split rather than to argue the bump away. `AxumError` is `pub use`d at the crate root and is the `Rejection` type for the extractors, so downstream code can match it exhaustively — adding a variant is a major change under Cargo's SemVer rules regardless of what the workspace itself does. That commit is preserved on `fix/320-axum-not-found` and will go up as a separate PR against the next minor.

I did check whether the 404 could be had in a patch by returning it as an `Ok(Response)` rather than an `Err`, which touches no public type. It works, but it makes the not-found case untyped and turns the eventual `AxumError::NotFound` into a second behaviour change, so splitting is the cleaner trade.

**Finding 3 is now a code change, not a changelog line.** Keeping `urlencoding = []` as a no-op is what makes this strictly non-breaking — without it, removing the implicit feature would itself have been a major change, so the patch would have been wrong for a second, less obvious reason.

**Finding 1 is fixed and I confirmed your reproduction.** `cargo check --all-targets -p authkestra-providers --no-default-features --features github,rustls-aws-lc-rs` fails today with unresolved `google` and `discord` imports. The tests now declare `required-features` and the job runs `--all-targets`. I verified the strengthened guard is load-bearing by dropping `required-features` from one test and watching that exact command fail.

**Finding 2** travels with the axum PR, bounded in both adapters as you preferred, so the two stay in parity.

PR #323, opened 2026-09-05
fix(axum): return 404 for an unknown OAuth provider (#320)

### LINKED ISSUE #320: Axum adapter returns 500 for an unknown OAuth provider; actix returns 404
## Summary

Requesting an OAuth provider that is not registered returns **500** from the axum adapter and **404** from actix, for the identical condition. The axum behaviour reports a client error as a server fault, and the two adapters disagree on the HTTP contract for the same request.

Found while generating a starter project from the playground: a session-only app with no providers registered returns 500 from `GET /auth/login/github`.

## The inconsistency

**axum** — `crates/authkestra-axum/src/helpers/api.rs`, three sites:

```rust
let flow: &Arc<dyn ErasedOAuthFlow> = match authkestra.providers.get(&provider) {
    Some(f) => f,
    None => return Err(AxumError::Internal("Provider not found".to_string())),
};
```

- `:287` in `axum_login_handler`
- `:329` in `axum_callback_handler`
- `:372` in `axum_callback_handler_stateless`

`AxumError::Internal` maps to `INTERNAL_SERVER_ERROR` (`:441`), and the enum has no not-found variant at all:

```rust
pub enum AxumError {
    Unauthorized(String),
    Internal(String),
    ComponentMissing(String),
}
```

**actix** — `crates/authkestra-actix/src/helpers/api.rs`, the three equivalent sites (`:170`, `:202`, `:347`):

```rust
None => return HttpResponse::NotFound().body(format!("Provider {provider} not found")),
```

404, and it names the provider.

## Why it matters

- **A 5xx for a client error.** A typo in a provider name is not a server fault. Anything alerting on 5xx rates sees noise it cannot act on, and a client cannot distinguish "you asked for something that does not exist" from "we broke".
- **The adapters disagree.** Swapping axum for actix, which the architecture otherwise makes cheap, silently changes the status code an application returns.
- The axum message also drops the provider name that actix includes, so the response says less.

## Reproducing

```rust
let engine = Engine::builder().session_store(store).build();  // no providers
let app = Router::new().merge(engine.axum_router());
```

```
GET /auth/login/github  ->  500 Internal Server Error
                            "Provider not found"
```

The same request against `axum_router_stateless`, and against `/auth/callback/{provider}`, behaves the same way.

## Suggested fix

Add a variant and use it at the three sites:

```rust
pub enum AxumError {
    Unauthorized(String),
    NotFound(String),
    Internal(String),
    ComponentMissing(String),
}

// in IntoResponse
AxumError::NotFound(msg) => (StatusCode::NOT_FOUND, msg),
```

```rust
None => return Err(AxumError::NotFound(format!("Provider {provider} not found"))),
```

That also brings the message into line with actix's.

Adding a variant to a public enum is technically breaking for anyone matching exhaustively on `AxumError`; `#[non_exhaustive]` would prevent that recurring, if that is a direction you want.

## Related, but separate

`AxumError::Internal(msg)` returns `msg` verbatim in the response body (`:441`). For genuine internal errors that risks putting internal detail in front of a caller — actix's own internal paths do the same (`ErrorInternalServerError(e.to_string())`). Worth a look, but it is a different concern from this one and I have not filed it.

Happy to send a PR for the status-code fix if useful.

marcjazz:
Fixes #320. **Split out of #322 because it cannot be a patch** — see below.

An unregistered OAuth provider returned 500 from the axum adapter and 404 from actix, so the two disagreed on the HTTP contract for the same request. `AxumError` had no not-found variant, so all three `providers.get()` misses fell back to `Internal`.

Adds `AxumError::NotFound`, maps it to 404, and uses it at all three sites in `crates/authkestra-axum/src/helpers/api.rs` (284, 328, 373).

### Why this is a minor, not a patch

Adding a variant to a public enum is a major change under [Cargo's SemVer reference](https://doc.rust-lang.org/cargo/reference/semver.html#enum-variant-new) unless the enum is `#[non_exhaustive]`. `AxumError` is `pub use`d at the crate root *and* is the `Rejection` type for the extractors, so downstream code can and plausibly does match it exhaustively. `BREAKING CHANGE:` is in the commit footer, so release-plz will take this to 0.9.0.

I deliberately did **not** add `#[non_exhaustive]`. It would prevent a recurrence, but it forecloses exhaustive matching for every downstream user from here on — that is your call to make, not something to fold into a bug fix. Adding it now would also be breaking in its own right, so it costs nothing extra to decide later.

### The reflected provider name

Review of #322 flagged that the 404 body echoes the provider name, which is an unvalidated, URL-decoded path segment — an unbounded attacker-controlled reflection where the previous code returned a constant. A 5000-character path came back as a 5019-byte body.

Not exploitable in a current browser, since the body is `text/plain; charset=utf-8` with no sniffing opportunity. Closed anyway, in **both** adapters: truncated at 64 characters with an ellipsis, each carrying a comment noting the other does the same. Fixing one and not the other would have quietly broken the parity that motivated the whole change.

Relocating the helper in actix surfaced a bug of my own making: it had landed under the `#[cfg(feature = "session")]` belonging to `handle_oauth_callback`, so the helper took the gate and the function lost it. Clippy's `empty_line_after_outer_attr` caught it; a `--no-default-features` build confirms the helper is not dead code where the features are off.

### Tests

Four new cases across the stateful and stateless routers, including one asserting a *registered* provider still redirects — a blanket 404 would satisfy the other assertions just as well.

Both new tests were verified load-bearing by reverting the fix they cover:

- the 404 test fails with `left: 500, right: 404`
- the truncation test fails with 5019 reflected bytes

That check mattered here: my first attempt at the 404 test **passed against the reverted fix**, because `sed` had not matched — rustfmt had wrapped the call sites across lines, so I was testing unmodified code. The first green run proved nothing.

Local gate green: fmt, clippy `-D warnings` across the workspace, and `cargo test --workspace --all-features`.

### DIFF
--- a/crates/authkestra-actix/src/helpers/api.rs
+++ b/crates/authkestra-actix/src/helpers/api.rs
@@ -18,14 +18,40 @@ use super::cookie::create_actix_cookie;
 /// finite. `authkestra-axum` bounds it identically — the two adapters return
 /// the same body by design, and a fix to one that skipped the other would
 /// quietly break that.
+///
+/// The budget is in **bytes**, not characters. What matters for a reflection is
+/// how much attacker-controlled data comes back, and 64 characters is up to 256
+/// bytes of UTF-8 — so a character bound does not actually bound the response.
+/// The cut is moved down to a `char` boundary so the result is still valid UTF-8.
 fn provider_for_display(provider: &str) -> String {
-    const MAX: usize = 64;
-    let shown: String = provider.chars().take(MAX).collect();
-    if shown.chars().count() < provider.chars().count() {
-        format!("{shown}\u{2026}")
-    } else {
-        shown
+    const MAX_BYTES: usize = 64;
+    if provider.len() <= MAX_BYTES {
+        return provider.to_string();
     }
+    let mut end = MAX_BYTES;
+    while !provider.is_char_boundary(end) {
+        end -= 1;
+    }
+    format!("{}\u{2026}", &provider[..end])
+}
+
+/// The 404 for an unregistered provider.
+///
+/// Built in one place because the `Content-Type` is load-bearing and easy to
+/// forget: `HttpResponseBuilder::body` sets no content type at all, and a
+/// response with none *and* an echoed path segment is sniffed as HTML by
+/// browsers — which turns this message into a reflected-XSS sink reachable from
+/// a plain link. `authkestra-axum` gets `text/plain` for free from
+/// `(StatusCode, String)`; actix does not, so it is set explicitly here, with
+/// `nosniff` as the belt to that braces.
+fn provider_not_found(provider: &str) -> HttpResponse {
+    HttpResponse::NotFound()
+        .content_type(actix_web::http::header::ContentType::plaintext())
+        .insert_header(("x-content-type-options", "nosniff"))
+        .body(format!(
+            "Provider {} not found",
+            provider_for_display(provider)
+        ))
 }
 
 #[derive(serde::Deserialize)]
@@ -186,10 +212,7 @@ pub async fn actix_login_handler<S, T>(
     let flow: &std::sync::Arc<dyn ErasedOAuthFlow> = match authkestra.providers.get(&provider) {
         Some(f) => f,
         None => {
-            return HttpResponse::NotFound().body(format!(
-                "Provider {} not found",
-                provider_for_display(&provider)
-            ));
+            return provider_not_found(&provider);
         }
     };
 
@@ -221,10 +244,7 @@ where
     let flow: &std::sync::Arc<dyn ErasedOAuthFlow> = match authkestra.providers.get(&provider) {
         Some(f) => f,
         None => {
-            return Ok(HttpResponse::NotFound().body(format!(
-                "Provider {} not found",
-                provider_for_display(&provider)
-            )));
+            return Ok(provider_not_found(&provider));
         }
     };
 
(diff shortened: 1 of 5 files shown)


marcjazz:
Review found a reflected-XSS sink this PR did not close, and I've pushed a fix in `356ab2b`.

**The 64-character bound was not sufficient on the actix side.** The PR reasons that the reflection is safe because the body is `text/plain; charset=utf-8`. That is true of axum, where `(StatusCode, String).into_response()` sets it. It is not true of actix: `HttpResponseBuilder::body` (actix-web 4.14, `src/response/builder.rs:291`) only calls `set_body` and never sets a content type. Verified against the real `actix_scope()` router on `397f985`:

```
GET /auth/login/%3Cimg%20src%3Dx%20onerror%3Dalert(1)%3E
STATUS=404  CONTENT_TYPE=None  NOSNIFF=None
BODY=Provider <img src=x onerror=alert(1)> not found
```

With no content type and no `nosniff`, browsers sniff the body as HTML — reflected XSS from a plain link, no CSRF token needed. Truncation does not help: that payload is 28 characters.

What changed:

- All three actix sites go through one `provider_not_found()` constructor setting `text/plain` and `nosniff`. One constructor rather than three call sites, because the content type is load-bearing and easy to omit — which is how this arose.
- `nosniff` on the axum side too, so both adapters carry the same guarantee instead of one relying on the tuple impl keeping its behaviour.
- `provider_for_display` now bounds **bytes** (64, cut to a char boundary) in both adapters. 64 chars is up to 256 bytes of UTF-8, so `body.len() < 200` in the axum test was asserting more than the code guaranteed.
- Two actix tests mirroring the axum ones — status, content type, `nosniff`, truncation. The cross-adapter parity both doc comments assert was enforced by nothing but the comments. Checked against a reverted fix: `an_unknown_provider_is_not_found_and_not_sniffable` fails without it.

Local gate: `fmt`, `clippy --workspace --all-features --all-targets -D warnings`, `test --workspace --all-features`, and `cargo check -p authkestra-actix --no-default-features --features rustls-aws-lc-rs` all green.

Not addressed, out of scope: axum returns 303 for a login redirect where actix returns 302. Pre-existing, and the same class of divergence #320 was filed about — worth its own issue.

PR #327, opened 2026-09-05
fix(actix): redirect with 303, matching axum

marcjazz:
Follow-up to #323. Reviewing that PR turned up a status-code divergence between the adapters that was out of its scope: `a_registered_provider_still_redirects` pins 303 on the axum side, while actix returned 302 for the identical request.

## Scope

Wider than the login redirect. **Six** actix sites diverged:

| | axum | actix (before) |
| --- | --- | --- |
| `GET /auth/login/{provider}` | 303 | 302 |
| `GET /auth/callback/{provider}` success | 303 | 302 |
| `GET /auth/logout` | 303 | 302 |
| OP `/authorize`, unauthenticated → `/login` | 303 | 302 |
| OP authorize redirect | 303 | 302 |
| OP consent path | 303 | 302 |

axum reaches all of these through `Redirect::to`, which is `SEE_OTHER`. actix used `HttpResponse::Found`.

## Why 303, and why actix moves

Every affected route is a `GET` on both sides, so 302 and 303 send the browser to the same place — no client behaviour changes in practice. 303 is the precise reading ("fetch the other resource with GET") without the method-rewriting that made 302 ambiguous, and it is what `Redirect::to` already picks. Converging there means one crate changes rather than two.

RFC 6749 §4.1.2 shows `302` in its authorization-response example. That is an example, not a requirement, and OIDC Core mandates no particular 3xx — so agreeing with the other adapter is worth more here than matching a non-normative snippet.

## Tests

Four actix assertions move from `FOUND` to `SEE_OTHER`, and the actix suite gains `the_login_redirect_is_a_see_other_like_axum`, mirroring axum's. Both are needed: a doc comment claiming the two adapters agree is not a thing that can fail, which is precisely how this drifted.

The `wired-endpoints` guide documented "HTTP 302 Redirect" and now says 303, noting the code is the same on either adapter.

## Not addressed

A site-by-site audit of the non-redirect statuses. actix reports errors through `ErrorUnauthorized` / `ErrorInternalServerError` where axum uses `AxumError` variants, and whether those line up condition for condition is a separate question. Worth its own issue if you want it swept.

## Gate

`fmt`, `clippy --workspace --all-features --all-targets -D warnings`, and `test --workspace --all-features` all green locally — 60 test binaries, same as before the change.

Note the `BREAKING CHANGE:` footer: the status code actix returns is observable, so a client asserting on 302 will see it change. 0.9.0 is already queued as breaking in #324.

### DIFF
--- a/crates/authkestra-actix/src/helpers/api.rs
+++ b/crates/authkestra-actix/src/helpers/api.rs
@@ -109,7 +109,12 @@ pub fn initiate_oauth_login_erased(
         .max_age(actix_web::cookie::time::Duration::minutes(15))
         .finish();
 
-    HttpResponse::Found()
+    // 303, not 302: the target is fetched with GET regardless of how this
+    // route was reached, which is exactly what a redirect-to-the-provider
+    // means. `authkestra-axum` sends 303 here too — `Redirect::to` is 303 —
+    // and the two adapters must not disagree on the status code for the same
+    // request. See issue #320 for the class of bug that causes.
+    HttpResponse::SeeOther()
         .insert_header((header::LOCATION, url))
         .cookie(cookie)
         .finish()
@@ -197,7 +202,7 @@ pub async fn handle_oauth_callback_erased(
         .success_url
         .unwrap_or_else(|| "/".to_string());
 
-    Ok(HttpResponse::Found()
+    Ok(HttpResponse::SeeOther()
         .insert_header((header::LOCATION, final_success_url))
         .cookie(cookie)
         .cookie(remove_cookie)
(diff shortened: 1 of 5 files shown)


marcjazz:
`codecov/patch` reports 50.00% of the diff hit (target 85.92%). That is exactly 3 of the 6 changed lines, and it is worth saying which 3.

The three OAuth client redirects — login, callback success, logout — are covered: the round-trip tests in both actix suites assert them, and this PR updates those assertions plus adds `the_login_redirect_is_a_see_other_like_axum`.

The three uncovered lines are the OpenID Provider redirects in `crates/authkestra-actix/src/op.rs`. Those handlers have **no adapter-level tests on either adapter** — `actix_authorize_handler` and `axum_authorize_handler` appear in no test file in either crate. So this is a pre-existing coverage gap that the diff walks through, not coverage this change removed. A one-line status change is also close to the least valuable place to spend the first OP test.

Closing it properly means standing up an `OpConfig` and a `CloneableOpStore` in the adapter suites, and doing it for one adapter only would create a fresh asymmetry of exactly the kind this PR exists to remove. That reads like its own issue — OP handler coverage in both adapters — rather than something to fold in here.

Flagging rather than silently merging past a red check.

PR #369, opened 2026-09-14
Gate facade features, make authkestra-engine optional (#325)

### LINKED ISSUE #325: The facade exposes fewer than half its sub-crates' features, and the engine is not gated at all
## Summary

The `authkestra` facade forwards 9 features. Its sub-crates offer roughly twice that. Everything a real application needs beyond sessions and OAuth — passkeys, TOTP, captcha, *any* persistent store, the macros, the OP server, device signatures — cannot be reached through the facade at all, so downstream crates have to depend on the sub-crates directly.

That defeats the point of having a facade. It should be an empty shell that pulls in code only when a feature turns it on.

## What the facade forwards

```
session  token  oidc  resource  axum  actix  github  google  discord
```

## What the sub-crates offer and the facade never exposes

| crate | unreachable through `authkestra` |
|---|---|
| `authkestra-engine` | `webauthn`, `totp`, `captcha`, `memory`, `redis`, `sql-postgres`, `sql-mysql`, `sql-sqlite` |
| `authkestra-axum` / `authkestra-actix` | `macros`, `op`, `captcha`, `devsig` |

So an app that wants passkeys writes:

```toml
authkestra-engine = { version = "0.8.1", features = ["webauthn", "session", "sql-sqlite"] }
authkestra-axum   = { version = "0.8.1", features = ["macros", "session"] }
```

and never mentions `authkestra` at all. That is the documented advice today, and it is what this playground does — see its ADR 0001, which exists solely because of this.

## The engine is not gated either

`crates/authkestra/Cargo.toml:14`:

```toml
authkestra-engine = { workspace = true, default-features = false, features = ["token"] }
```

Every other dependency is `optional = true`. This one is not, and it hardcodes `token`. Consequences:

- `authkestra` with `--no-default-features` still compiles the engine, with `token` on. There is no way to get an empty facade.
- The facade's own `token = ["authkestra-engine/token"]` feature is therefore a no-op — it is always enabled, and `#[cfg(feature = "token")]` at `src/lib.rs:17` gates a re-export of something that is unconditionally compiled.
- There is no way to use the facade *without* the token machinery, even for a pure session app.

## Suggested shape

Make the engine optional like everything else, and forward the rest:

```toml
[dependencies]

[features]
# storage
memory       = ["authkestra-engine?/memory"]
redis        = ["authkestra-engine?/redis"]
sql-sqlite   = ["authkestra-engine?/sql-sqlite"]
sql-postgres = ["authkestra-engine?/sql-postgres"]
sql-mysql    = ["authkestra-engine?/sql-mysql"]

# methods
webauthn = ["authkestra-engine?/webauthn"]
totp     = ["authkestra-engine?/totp"]
captcha  = ["authkestra-engine?/captcha", "authkestra-axum?/captcha", "authkestra-actix?/captcha"]

# adapters
macros = ["authkestra-axum?/macros", "authkestra-actix?/macros"]
op     = ["authkestra-axum?/op", "authkestra-actix?/op"]
devsig = ["authkestra-axum?/devsig", "authkestra-actix?/devsig"]
```

with the matching `#[cfg(feature = ...)]` re-exports.

## Why this is cheap

**It is not a breaking change.** Adding features is a minor bump under Cargo's SemVer rules, and the optional-dependency plumbing is already in place for five of the six crates. It is mostly passthrough lines.

The one judgement call is making `authkestra-engine` optional: code that relies on the facade re-exporting engine types with no features named would stop resolving. A `default = ["engine"]` (or keeping the current behaviour under `default`) preserves that while letting `--no-default-features` mean what it says.

## Worth a test

A feature that is silently always-on, like `token` today, is invisible without one. Something along the lines of the `provider-features` matrix in #322:

```sh
for f in webauthn totp captcha memory redis sql-sqlite macros op devsig; do
  cargo check -p authkestra --no-default-features --features "$f,rustls-aws-lc-rs" || exit 1
done
```

## Context

Found while building a starter-kit generator that emits projects for arbitrary combinations of authkestra features. Every generated project has to depend on sub-crates directly, and the README has to explain why — which reads as a wart in the framework rather than a choice.

marcjazz:
## Summary

Make `authkestra-engine` an optional dependency and properly gate all facade features. Previously, `--no-default-features` still compiled the engine with the token feature enabled, and features like webauthn, totp, storage backends, and adapter extras (macros, op, devsig) couldn't be reached through the facade.

## Changes

- Make `authkestra-engine` optional in `[dependencies]`
- Add `engine` feature that enables the engine and token support (included in default)
- Require storage backends and auth methods to enable the engine
- Gate core engine re-exports (core, flow, Authkestra type) on engine feature
- Add re-exports for webauthn and totp when their features are enabled
- Update TLS backend features to use optional dependency syntax
- Add CI job to verify each facade feature compiles in isolation

## Test plan

All facade features (webauthn, totp, captcha, memory, redis, sql-sqlite, sql-postgres, sql-mysql, macros, op, devsig) compile in isolation with `--no-default-features`.

With these changes:
- `authkestra --no-default-features` disables the engine entirely
- Individual features like `webauthn`, `memory`, `redis`, etc. work standalone
- Default behavior unchanged (engine enabled by default)
- All storage and auth methods are now accessible through the facade

Closes #325


marcjazz:
Follow-up commit: `session` and `token` were missing `engine` as an implied feature (every other feature needing the engine — memory/redis/sql-*/webauthn/totp/captcha — got it, these two didn't). `--no-default-features --features session` failed to compile (`authkestra_engine::store` unresolved) because the `?/` optional-forwarding syntax never activates `dep:authkestra-engine` on its own. Fixed, added `session`/`token` to the facade-features CI matrix so it's covered going forward, and re-verified all 13 features plus the empty-facade and default cases compile.

marcjazz:

```toml
[dependencies]

[dev-dependencies]

[features]

# Core features
```

PR #375, opened 2026-09-14
fix(engine)!: remove authkestra-engine's no-op session feature

marcjazz:
## Summary

- Follow-up to #373's no-op/feature-gate audit. `authkestra-engine`'s `session` Cargo feature never gated anything in its own source — re-verified independently: `grep -rn 'feature = "session"' crates/authkestra-engine/src/` returns nothing. Session-store functionality is actually gated by `memory`/`redis`/`sql-postgres`/`sql-mysql`/`sql-sqlite`, not `session`.
- Removed the dead `session = []` line from `crates/authkestra-engine/Cargo.toml`.
- Fixed every downstream `Cargo.toml` that forwarded to it (`authkestra-engine/session`), so the workspace doesn't hard-error on a dangling feature reference:
  - `crates/authkestra/Cargo.toml` (facade, both the `[dependencies]`/`[dev-dependencies]` feature list and the `session = [...]` feature forward)
  - `crates/authkestra-axum/Cargo.toml`
  - `crates/authkestra-actix/Cargo.toml`
- The facade/axum/actix crates keep their **own** `session` feature — confirmed via grep that it really does gate real code in each (`crates/authkestra/src/lib.rs`, `crates/authkestra-axum/src/lib.rs` + `helpers/`, `crates/authkestra-actix/src/lib.rs` + `helpers/`) — only the dead forward to `authkestra-engine` is dropped.

## Breaking change

`fix(engine)!:` — anyone building `authkestra-engine` directly with `--features session` now gets a Cargo "unknown feature" error instead of a silent no-op. This costs nothing functionally: the feature never compiled anything in or out of the engine. Consumers going through the `authkestra`/`authkestra-axum`/`authkestra-actix` facades are unaffected since their own `session` features are unchanged.

## Verification

- `cargo check -p authkestra-engine` (default features, and combos with `memory`/`redis`/`totp`/`webauthn`/`sql-postgres`/`sql-mysql`/`sql-sqlite`/`captcha`) — all pass.
- `cargo check -p authkestra-engine --features session` now correctly errors with "the package 'authkestra-engine' does not contain this feature: session".
- `cargo check -p authkestra-axum` / `authkestra-actix` with `session` alone, and both with `--all-features` — pass.
- `cargo check -p authkestra --features full` — passes.
- `cargo test -p authkestra-engine` (with memory/redis/totp/webauthn/sql-sqlite) — 189 + integration + doc tests pass.
- `cargo test -p authkestra-axum --all-features` and `cargo test -p authkestra-actix --all-features` — all pass, including the `engine_session_integration` suites.
- `cargo test -p authkestra --features full` — passes.

### Pre-existing, unrelated issue found (not touched here)

While isolating single features on `authkestra-axum`/`authkestra-actix`, several feature combos that don't include the crate's own `session` feature (e.g. `token` alone, `resource` alone, and `authkestra-actix`'s **default** features) fail to compile due to a `#[cfg(feature = "session")]`-gated import (`FromRequestParts`) that other, differently-gated code relies on unconditionally. Verified with `git checkout` back to the pre-change `Cargo.toml`s that this reproduces identically on unmodified `main` — it predates and is unrelated to this change. Flagging it since it affects the feature-isolation matrix, but it's out of scope for this PR (no dead `authkestra-engine/session` forward involved).

## Test plan

- [x] `cargo check -p authkestra-engine` (default + feature combos)
- [x] `cargo check -p authkestra-engine --features session` fails as expected
- [x] `cargo check -p authkestra-axum` / `authkestra-actix` (session, all-features)
- [x] `cargo check -p authkestra --features full`
- [x] `cargo test -p authkestra-engine` / `authkestra-axum` / `authkestra-actix` / `authkestra`


marcjazz:
Flagging for whoever merges these: this and **#369** will conflict on `crates/authkestra/Cargo.toml`'s `session` line. #369 has `session = ["engine", "authkestra-engine?/session"]` (fixes `session` to imply `engine`, but still forwards to a feature this PR deletes); this PR has `session = []` (drops the dead forward correctly, but predates #369 so it doesn't imply `engine`). Whichever merges second needs to reconcile to `session = ["engine"]` — neither branch alone has that final correct line. I'll fix it when the second one merges unless someone gets there first.

marcjazz:

```toml
[dependencies]

[dev-dependencies]

[features]

# Core features
```

Issue #370, opened 2026-09-14
Implement a GNAP (RFC 9635) authorization server prototype behind a `gnap` feature flag


marcjazz:
## Context

`docs/rfc-004-gnap-flow.md` is the design RFC for GNAP (RFC 9635) support. Its
§6 lays out a staged G0–G6 plan for an actual GNAP authorization server
prototype in `authkestra-op`, and §11 (added by the PR that closes this issue's
predecessor task) records that the `Flow` trait itself has now been updated to
be GNAP-*shaped* (a defaulted `Flow::execute_with_parts` carrying `RequestParts`,
an optional JSON `body` on `FlowContext`, and a `FlowResult::Document` variant) —
but no GNAP protocol logic exists anywhere in the workspace yet.

This issue tracks that remaining, much larger body of work, deliberately kept
out of the trait-update PR so that PR stayed reviewable.

## Scope (RFC-004 §6)

- **G0** — `gnap` feature flag (off by default) in `authkestra-op`,
  `authkestra-axum`, `authkestra-actix`, and the `authkestra` facade; the §2/§3
  request and response types with serde round-trip tests against the literal
  JSON examples in RFC 9635 Appendix B; `OPTIONS /gnap` discovery endpoint.
- **G1** — `GnapGrantStore` trait + in-memory implementation, mirroring the
  existing `OpStore` pattern. Continuation-token lookup and single-use
  `interact_ref` consumption must be atomic (RFC-004 §6.3).
- **G2** — software-only grant (no interaction): `POST /gnap` with
  `client.key`, straight to `approved`, single access token issued via the
  existing `TokenManager`. First end-to-end runnable path; carries the
  required facade example (`axum_gnap_grant.rs`), per AGENTS.md "Examples
  Live in the Facade." Actix routes wired in the same PR (AGENTS.md
  "Framework Wiring" DoD); the actix example is deferred until the feature
  stabilises (RFC-004 §6.3).
- **G3** — `user_code` interaction + polling, reusing the existing device-flow
  verification handler and `DeviceCodeStore` shape.
- **G4** — `redirect` start + `redirect` finish, including the §4.2.3
  interaction hash (known-answer test is specified in RFC-004 §6.5).
- **G5** — multiple labelled access tokens (§2.1.2 / §3.2.2).
- **G6** — revisit RFC-004 §5's Route A vs Route B decision (does GNAP go
  through `Flow`, or live as its own `authkestra-op` trait?) with a working
[...]

PR #377, opened 2026-09-14
fix(engine): de-feature token — always compile token machinery in

### LINKED ISSUE #376: authkestra-engine: the token feature is default-on but not genuinely optional
## What's broken

`authkestra-engine`'s `token` Cargo feature is default-on but was not actually optional. Building it explicitly without token support:

```
cargo check -p authkestra-engine --no-default-features --features rustls-aws-lc-rs
```

fails with:

```
error[E0392]: type parameter `T` is never used
  --> crates/authkestra-engine/src/engine.rs:47:32
   |
47 | pub struct Engine<S = Missing, T = Missing> {
   |                                ^^^^^^^^^^^ unused type parameter

error[E0392]: type parameter `T` is never used
   --> crates/authkestra-engine/src/engine.rs:104:39
    |
104 | pub struct EngineBuilder<S = Missing, T = Missing> {
    |                                       ^^^^^^^^^^^ unused type parameter

error[E0282]: type annotations needed
  --> crates/authkestra-engine/src/engine.rs:90:9
   ...cannot infer type of the type parameter `T` declared on the struct `EngineBuilder`
```

Root cause: `Engine<S, T>`/`EngineBuilder<S, T>`'s `token_manager: T` field was behind `#[cfg(feature = "token")]`, so when the feature was off the field disappeared and `T` became a structurally unused type parameter — plus the `Missing`-typed initializers lost their inference anchor.

Every downstream crate (`authkestra-axum`, `authkestra-actix`, `authkestra-oidc`, `authkestra-op`, `authkestra-providers`, `authkestra-resource`, the `authkestra` facade) worked around this by hardcoding `authkestra-engine = { ..., features = ["token"] }` unconditionally on their base dependency line, rather than forwarding their own `token` feature to `authkestra-engine/token` the way every other feature is forwarded. So nothing broke in practice — but the flag didn't mean what it claimed, and anyone reaching for `authkestra-engine --no-default-features` directly (e.g. to check what a reduced build looks like, as done while investigating #373) hit a confusing compiler error instead of a working reduced build.

Found while investigating #373. This is also what #369 (the facade's feature-forwarding work) was unable to be fully honest about: it could not forward `token` conditionally like it does every other feature, because a real "token off" build of the engine didn't compile.

## Investigation

Read `crates/authkestra-engine/src/engine.rs`'s typestate builder (`Missing`/`Configured<T>`, mirroring the pattern the README's Quick Start shows for `S`) and `crates/authkestra-engine/src/token/mod.rs`, plus `crates/authkestra-engine/Cargo.toml`'s `[dependencies]`. Two things stood out:

1. **The typestate slot itself was already well-formed as a plain generic field.** `S` (the session-store slot) is *not* cfg-gated and works fine as `Missing` by default — `T` should behave identically once its field isn't cfg'd out. `aliases.rs` already assumed this: `AkWebAppEngine = Engine<Configured<Arc<dyn SessionStore>>, Missing>` is defined unconditionally, so the crate's own type aliases already expected `T = Missing` to type-check with no feature required.

2. **The `token` feature never actually gated anything below the typestate layer.** `pub mod token;` in `lib.rs` is unconditional (not `#[cfg(feature = "token")]`), and none of its dependencies (`jsonwebtoken`, `aes-gcm`, `rsa`, `ed25519-dalek`, `tokio` with `"full"`) are `optional = true` in `Cargo.toml` — they're all compiled in regardless of the feature. Worse, `jsonwebtoken` is used unconditionally by `Engine::authenticate()` itself (`crates/authkestra-engine/src/engine.rs`, MFA continuation-token signing/verification) — a core, always-on code path with no relationship to the `token` feature. So disabling `token` never removed a single dependency, line of compiled code, or byte of binary size; the feature only ever gated a thin accessor layer on `Engine`/`EngineBuilder` (`token_manager()`, `jwt_secret()`, `issue_token()`, `TokenManagerState`/`HasTokenManager`) — and gated it incorrectly, per the E0392 above.

## Direction chosen: (b) — de-feature `token`

I considered making `T` a well-formed `PhantomData`-style marker when `token` is off (option a), which would have made `--no-default-features` compile. But given finding (2), that would have been dishonest in the other direction: a "token disabled" build would still pull in the full JWT/AES/RSA/Ed25519 stack and still ship `TokenManager`, `token::dpop`, `token::sd_jwt`, etc. — the feature would compile cleanly while promising a reduction it can't deliver, since the core MFA flow already depends on `jsonwebtoken` unconditionally. There's no way to make token support *genuinely* optional without also carving MFA's continuation-token signing out from under it, which is a materially larger, separate change.

Given the feature was never wired to anything removable, the honest fix is (b): remove `token` as an optional Cargo feature and always compile the token machinery in. This is exactly what every downstream crate's hardcoded `features = ["token"]` was already assuming in practice — it was already telling the truth; this PR just makes the manifests agree with it.

## What this PR does

- `authkestra-engine`: drops `token` from `[features]`/`default`; removes every `#[cfg(feature = "token")]` in `engine.rs` and `aliases.rs` so `Engine<S, T>`'s token-manager slot, `TokenManagerState`/`HasTokenManager`, and the `AkApiEngine`/`AkEngine` aliases are always present.
- Downstream `Cargo.toml`s (`authkestra-axum`, `authkestra-actix`, `authkestra-oidc`, `authkestra-op`, `authkestra-providers`, `authkestra-resource`, `authkestra`): drop the now-nonexistent `"token"` entry from their `authkestra-engine = {...}` dependency declarations. `authkestra-axum`/`authkestra-actix`/`authkestra`'s own local `token` feature (which gates *their own* token-issuing wrapper code, e.g. `issue_token` endpoints) is kept as a feature name — since removing it cascades into `required-features` on several `[[test]]` targets and composite features like `op`/`full` — but it no longer forwards to `authkestra-engine/token` (nothing to forward to); it's now a no-op alias, called out with a comment at each site.

## Deferred to a follow-up

Whether `authkestra-axum`/`authkestra-actix`/`authkestra`'s own `token` feature should be collapsed away entirely (since the thing it used to conditionally forward to `authkestra-engine` is now unconditional, so on its own it now toggles nothing at the engine level) is left open. Doing that touches `required-features` on `[[test]]` targets, the `op`/`full` composite features, and downstream `#[cfg(feature = "token")]` gates on wrapper functions in `authkestra-axum`/`authkestra-actix` — a larger, separate surface than this fix, mirroring how #371 filed #370 for GNAP's remaining work instead of doing it all in one PR.

Refs #373, #369.

marcjazz:
## Summary

`authkestra-engine`'s `token` Cargo feature was default-on but not genuinely optional — building it with `--no-default-features --features rustls-aws-lc-rs` failed with E0392/E0282 because `Engine<S, T>`/`EngineBuilder<S, T>`'s `token_manager: T` field was `#[cfg(feature = "token")]`-gated, leaving `T` structurally unused when the feature was off. Every downstream crate worked around this by hardcoding `features = ["token"]` unconditionally on its `authkestra-engine` dependency instead of forwarding it, so nothing broke in practice, but anyone building the engine standalone hit a confusing compile error instead of a working reduced build.

Full investigation and decision reasoning are in #376.

**Chosen direction: de-feature `token` (option b), not make it a well-formed no-op (option a).** The feature never actually gated anything below the typestate layer: `pub mod token;` is unconditional, none of its dependencies (`jsonwebtoken`, `aes-gcm`, `rsa`, `ed25519-dalek`, `tokio "full"`) are `optional = true`, and `jsonwebtoken` is already used unconditionally by `Engine::authenticate()`'s MFA continuation-token flow. Disabling `token` never removed a single dependency or line of compiled code — it only gated a thin (and broken) accessor layer on `Engine`. Making it a clean `PhantomData`-style no-op would have compiled, but would have kept promising a reduced build that the feature could never actually deliver. Downstream's hardcoded `features = ["token"]` was already telling the truth; this PR makes the manifests agree with it.

## Changes

- `authkestra-engine`: dropped `token` from `[features]`/`default`; removed every `#[cfg(feature = "token")]` in `engine.rs`/`aliases.rs` — the token-manager typestate slot, `TokenManagerState`/`HasTokenManager`, and the `AkApiEngine`/`AkEngine` aliases are now always present.
- `authkestra-axum`, `authkestra-actix`, `authkestra-oidc`, `authkestra-op`, `authkestra-providers`, `authkestra-resource`, `authkestra` (facade): dropped the now-nonexistent `"token"` entry from their `authkestra-engine = {...}` dependency declarations. `authkestra-axum`/`authkestra-actix`/`authkestra`'s own local `token` feature (gating their own token-issuing wrapper code, and referenced by several `required-features`/composite features) is kept as a name but no longer forwards to `authkestra-engine/token` — it's now a no-op, called out with a comment at each site.

## Deferred

Whether downstream's own now-meaningless `token` feature should be collapsed away entirely is left open — that touches `required-features` on several `[[test]]` targets and composite features (`op`, `full`) in `authkestra-axum`/`authkestra-actix`/`authkestra`, a larger surface than this fix. Noted in #376, mirroring how #371 filed #370 for GNAP's remaining work rather than doing it all in one PR.

## Test plan

- [x] `cargo check -p authkestra-engine --no-default-features --features rustls-aws-lc-rs` now compiles clean (previously E0392/E0282)
- [x] Feature-isolation matrix on `authkestra-engine`: `session`, `webauthn`, `totp`, `captcha`, `memory`, `redis`, `sql-postgres`, `sql-mysql`, `sql-sqlite` (each alone with `rustls-aws-lc-rs`), plus default, `--all-features`, and `rustls-no-provider`
- [x] `cargo check --workspace --all-targets`
- [x] `cargo test -p authkestra-engine` (156 tests + doctests)
- [x] `cargo test` for every downstream crate whose `Cargo.toml` changed: `authkestra-axum --all-features`, `authkestra-actix --all-features`, `authkestra-oidc`, `authkestra-op`, `authkestra-providers --all-features`, `authkestra-resource --all-features`, `authkestra-facade-tests`, `authkestra`

Closes #376. Refs #373, #369.


marcjazz:

```toml
[dependencies]

[dev-dependencies]

[features]

# Core features
```

PR #384, opened 2026-09-14
refactor(facade)!: one name for the engine, and drop the two empty features

marcjazz:
Raised by the maintainer while reading `crates/authkestra/src/lib.rs`. Four related problems in the facade's public surface, all of the same kind: names and features that claim something they no longer deliver.

## 1 — `authkestra_engine` was exported under three names (`1ccf67a4`)

```rust
pub use authkestra_engine as core;
pub use authkestra_engine as flow;      // same crate
pub use authkestra_engine as token;     // same crate again
pub use authkestra_engine::store;                 // gated on `session`
pub use authkestra_engine::store as persistence;  // same module, wider gate
```

`flow` and `token` are leftovers from RFC-001: `authkestra-flow` and `authkestra-token` were real crates before being merged into `authkestra-engine`, and the facade kept their names so downstream paths survived the merge. The aliases outlived the crates.

They were misleading, not merely redundant — `authkestra::flow::` handed you the *entire* engine including token and session machinery, and `token` was gated on `feature = "token"` while re-exporting far more than that feature covers.

`persistence` was a mistake in #369 last month (`git log -S` pins it to `efdb50a3`): the same module as `store`, under a *wider* gate, so which name resolved depended on which feature you enabled.

**Kept `core`** — usage decided it: 15 call sites, against `flow`'s single doc comment and zero for `token`/`store`/`persistence`. Everything removed stays reachable through it (`authkestra::core::store`, `::token`, `::flow`).

## 2 — the `session` and `token` features were empty (`065e84e5`)

Both stopped forwarding anything once the engine dropped its own `session` feature (#375) and made token machinery unconditional (#377). With the duplicate re-exports gone above, they gated nothing at all — the same "feature that turns nothing on" class this release already spent several PRs auditing away.

Two traps, neither visible from the feature table:

- **`full` reached `engine` only through `session`/`token`.** Nothing else in that list pulls the engine in, so dropping them without naming `engine` explicitly would have quietly removed `authkestra::core` and `Authkestra` from a `--features full` build.
- **`authkestra-facade-tests` hard-depends on both**, and Cargo fails *resolution* rather than compilation on an unknown feature — so it surfaces as `failed to select a version for authkestra`, with nothing pointing at the cause.

Also corrects a comment from #377's conflict resolution claiming the `axum`/`actix` composites name the facade's `session`/`token`. They forward to `authkestra-axum/session` — the *adapter's* features, which are real and unaffected.

## 3 — `full` named 7 of 19 features (`190e3f23`)

It omitted `webauthn`, `totp`, `captcha`, all five storage backends, `resource`, `macros`, `op` and `devsig`. #369 added twelve facade features and updated `full` for none of them, so a caller asking for "full" got no passkeys, no TOTP and no OP server, with nothing indicating anything was missing.

Fixing only the instance would leave the next one to the same discipline that already failed, so **`authkestra-facade-tests` now parses the facade's own manifest** and asserts `full` stays exhaustive. Verified against the real defect: deleting `"webauthn"` from `full` fails the test naming `webauthn`.

## 4 — but the storage backends are excluded, by decision (`0d3d1689`)

An application runs one store; naming all five in `full` put a Redis client and three SQL drivers into every build for four things that go unused. `full` now means **"every way of authenticating"**, with the store named alongside it: `features = ["full", "sql-postgres"]`.

Because the previous commit's whole argument was that a `full` missing things is indistinguishable from a `full` that drifted, the exclusion is **declared** — `EXCLUDED_FROM_FULL` carries the backends and the reasoning, kept separate from `NOT_CAPABILITIES` (the aggregates and the TLS backend *choice*) since "not a capability" and "a capability we deliberately don't enable" are different claims. Two further guards: `full` and the exclusion list may not contradict each other, and a stale exclusion must be dropped rather than silently exempting a future feature that reuses the name.

All four guards verified against real defects, not assumed: removing `"totp"` from `full` fails naming `totp`; adding `"redis"` to `full` fails naming `redis`.

## Resulting surface

```rust
pub use authkestra_engine as core;                    // the one name for the engine
pub type Authkestra<S, T> = authkestra_engine::Engine<S, T>;
pub use authkestra_engine::store;                     // short path, gated on storage backends
```

plus the genuinely distinct sub-crate re-exports (`oidc`, `axum`, `actix`, `op`, `webauthn`, `totp`, …).

## Breaking

- `authkestra::flow`, `authkestra::token`, `authkestra::persistence` removed — use `authkestra::core` or `authkestra::store`.
- The facade's `session` and `token` **features** removed — use `engine`, or any storage backend, which implies it. The identically-named features on `authkestra-axum`/`authkestra-actix` are unaffected.
- `authkestra::store` now requires a storage backend feature rather than `session`.
- `full` enables considerably more than before (every auth capability), and still no storage backend.

## Verification

Every facade feature in isolation, plus `full`, `engine`, empty and default; `cargo test --workspace --all-features`; `check --workspace --all-features`; `clippy -D warnings`; `fmt --check`; the `authkestra-facade-tests` suites. The CI facade-features matrix listed `session`/`token` (added earlier this same release cycle) and now checks `engine`.

Targets 0.11.0, alongside the other breaking changes already in that release.


marcjazz:
Second commit (`065e84e5`) removes the facade's `session` and `token` features, per follow-up. Both had stopped forwarding anything after #375/#377, and with the duplicate re-exports gone in the first commit they gated nothing at all — the same "feature that turns nothing on" class this release already spent several PRs auditing away.

Two traps this had to avoid, neither visible from the feature table:

- **`full` reached `engine` only through `session`/`token`.** Nothing else in that list pulls the engine in, so dropping them without naming `engine` explicitly would have quietly removed `authkestra::core` and `Authkestra` from a `--features full` build. `full` now names `engine` directly.
- **`authkestra-facade-tests` hard-depends on both.** Cargo fails *resolution*, not compilation, on an unknown feature — so this shows up as `failed to select a version for authkestra` with nothing pointing at the cause. Switched to `engine`.

Also corrects a comment I wrote during #377's conflict resolution, which claimed the `axum`/`actix` composite features name the facade's `session`/`token`. They forward to `authkestra-axum/session` — the *adapter's* features, which are real and unaffected. Only `full` referenced the facade's own. And the CI facade-features matrix listed both (added earlier in this same release cycle); it now checks `engine`.

Verified: `full`, `engine`, empty, default and every storage/method/adapter feature in isolation; `session`/`token` now correctly fail as unknown features; `cargo test --workspace --all-features`, `check --workspace --all-features`, `clippy -D warnings`, `fmt --check` all clean; `authkestra-facade-tests` suites pass.

